PR #116, opened 2021-09-21
Pull docker images from harbor registry for e2e test

jamiemonserrate:
Otherwise we are running into docker rate limits


anusha94:
@jamiemonserrate I don't think we need this PR anymore. I remember having this workaround when we were using Docker Cluster for e2e tests and pulling images from dockerhub resulted in rate limit errors.
Does it make sense to close this PR?

PR #205, opened 2021-11-03
Add support for bundle registry and tag fields

### LINKED ISSUE #203: Support bundleRepository and bundleTag fields for BYOH bundles
- user must be able to specify the bundle repository and the bundle tag for a given BYOH bundle
- any new fields that need to be communicated to the agent should be via byoh CR
- write unit tests wherever applicable

anusha94:

- introduce bundleRegistry and bundleTag fields on ByoCluster spec
- add annotations on host for registry and tag
- unit tests for the byomachine controller

Fixes #203

### DIFF
--- a/apis/infrastructure/v1beta1/byocluster_types.go
+++ b/apis/infrastructure/v1beta1/byocluster_types.go
@@ -20,6 +20,14 @@ type ByoClusterSpec struct {
 	// ControlPlaneEndpoint represents the endpoint used to communicate with the control plane.
 	// +optional
 	ControlPlaneEndpoint APIEndpoint `json:"controlPlaneEndpoint"`
+
+	// BundleLookupBaseRegistry is the base Registry URL that is used for pulling byoh bundle images,
+	// if not set, the default will be set to https://projects.registry.vmware.com/cluster_api_provider_bringyourownhost
+	// +optional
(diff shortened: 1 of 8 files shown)


yixingjia:
Copy the slack discussion here. 
Overall I'm good with this changes but with some comments below. 

Is the bundle tag a required filed?
Maybe we can follow the kubeadm  behavior ,If bundle  tag is not set we should automatically generate the bundle tag according to the Kubernates version + OS information.  just a suggestion.
We also need to update the template as well to reflect this changes.
This also inspire me to thought more about the default agent bundle behavior. Seems we will not support provision K8s cluster if we don’t have the bundle even if os already have kubeadm/kubelet/kubectl ready on the OS, right?

yixingjia (approved):

dharmjit:
> Copy the slack discussion here. Overall I'm good with this changes but with some comments below.
> 
> Is the bundle tag a required filed? Maybe we can follow the kubeadm behavior ,If bundle tag is not set we should automatically generate the bundle tag according to the Kubernates version + OS information. just a suggestion. We also need to update the template as well to reflect this changes. This also inspire me to thought more about the default agent bundle behavior. Seems we will not support provision K8s cluster if we don’t have the bundle even if os already have kubeadm/kubelet/kubectl ready on the OS, right?

@yixingjia, As discussed we will proceed with this PR and we might need to look at this holistically at a later point in time and how UX can be improved for the same.

PR #207, opened 2021-11-03
Installer OS mapping

ratanasovvmw:
A BYOH bundle and installer Algo are developed and tested using a specific OS version (including patch version).
This is why the patch version is included in the BYOH bundle name. An old BYOH Bundle may work with newer OS patches and versions.

This PR make the installer smart. It can map any OS version to any BYOH Bundle and installer Algo.

Registry contains
1. Entries associating BYOH Bundle in the Repository with Installer in Host Agent
2. Entries that match an OS to a BYOH Bundle OS in the Repository

cli --list-supported
```
OS                      K8S Version     BYOH Bundle Name
---                     -----------     ----------------
Ubuntu_20.04.*_x86-64    v1.22.1        byoh-bundle-ubuntu_20.04.1_x86-64_k8s_v1.22.1
```

Testing done
On a Ubuntu 18.04
./cli --list-supported
./cli --preview-os-changes --os Ubuntu_20.04.*_x86-64 --k8s v1.22.1
./cli --install --os Ubuntu_20.04.1_x86-64 --k8s v1.22.1 --bundle-repo <REPO>
./cli --uninstall --os Ubuntu_20.04.1_x86-64 --k8s v1.22.1 --bundle-repo <REPO>

### DIFF
--- a/agent/installer/cli.go
+++ b/agent/installer/cli.go
@@ -69,7 +69,6 @@ func Main() {
 
 func listSupported() {
 	w := new(tabwriter.Writer)
-	// minwidth, tabwidth, padding, padchar, flags
 	const (
 		minwidth = 8
 		tabwidth = 8
--- a/agent/installer/cli.go
+++ b/agent/installer/cli.go
@@ -6,10 +6,10 @@ package installer
 import (
 	"flag"
 	"fmt"
-	"os"
-	"text/tabwriter"
 	"github.com/go-logr/logr"
 	"k8s.io/klog/klogr"
+	"os"
+	"text/tabwriter"
 )
 
 var (
@@ -70,36 +70,35 @@ func Main() {
 func listSupported() {
 	w := new(tabwriter.Writer)
 	// minwidth, tabwidth, padding, padchar, flags
-	w.Init(os.Stdout, 8, 8, 0, '\t', 0)
+	const (
+		minwidth = 8
+		tabwidth = 8
+		padding  = 0
+		flags    = 0
+	)
+	w.Init(os.Stdout, minwidth, tabwidth, padding, '\t', flags)
 	defer w.Flush()
 
-	fmt.Fprintf(w, "%s\t%s\t%s\n", "OS",  "K8S Version", "BYOH Bundle Name")
+	fmt.Fprintf(w, "%s\t%s\t%s\n", "OS", "K8S Version", "BYOH Bundle Name")
 	fmt.Fprintf(w, "%s\t%s\t%s\n", "---", "-----------", "----------------")
 
-	osList, aliasMap := ListSupportedOS()
-
-	for _, os := range osList {
-		for _, k8s := range ListSupportedK8s(os) {
-			fmt.Fprintf(w, "%s\t %s\t%s\n", os, k8s, GetBundleName(os, k8s))
+	osFilters, osBundles := ListSupportedOS()
+	for i := range osFilters {
+		for _, k8s := range ListSupportedK8s(osBundles[i]) {
+			fmt.Fprintf(w, "%s\t %s\t%s\n", osFilters[i], k8s, GetBundleName(osBundles[i], k8s))
 		}
-	}
-
-	for a, o := range aliasMap {
-		for _, k8s := range ListSupportedK8s(o) {
-                        fmt.Fprintf(w, "%s\t %s\t%s\n", a, k8s, GetBundleName(o, k8s))
-                }
 	}
 }
 
 func detectOS() {
 	osd := osDetector{}
-	os, err := osd.Detect()
+	detectedOs, err := osd.Detect()
 	if err != nil {
 		klogger.Error(err, "Error detecting OS")
 		return
 	}
 
-	fmt.Printf("Detected OS as: %s", os)
+	fmt.Printf("Detected OS as: %s", detectedOs)
 }
 
 func runInstaller(install bool) {
--- a/agent/installer/cli.go
+++ b/agent/installer/cli.go
@@ -6,13 +6,14 @@ package installer
 import (
 	"flag"
 	"fmt"
+	"os"
+	"text/tabwriter"
 	"github.com/go-logr/logr"
 	"k8s.io/klog/klogr"
 )
 
 var (
-	listSupportedFlag    = flag.Bool("list-supported", false, "List all supported OS and Kubernetes versions")
-	listBundlesFlag      = flag.Bool("list-bundles", false, "List the BYOH Bundle names for all supported OS and Kubernetes versions")
+	listSupportedFlag    = flag.Bool("list-supported", false, "List all supported OS, Kubernetes versions and BYOH Bundle names")
 	detectOSFlag         = flag.Bool("detect", false, "Detects the current operating system")
 	installFlag          = flag.Bool("install", false, "Install a BYOH Bundle")
 	uninstallFlag        = flag.Bool("uninstall", false, "Unnstall a BYOH Bundle")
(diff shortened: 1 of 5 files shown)


agent/installer/installer.go
```diff
@@ -37,21 +37,28 @@ func getSupportedRegistry(bd *bundleDownloader, ob algo.OutputBuilder) registry
 		os   string
 		k8s  string
 		algo algo.K8sStepProvider
+		osFilter string
 	}{
-		{"Ubuntu_20.04.1_x86-64", "v1.22.1", &algo.Ubuntu20_4K8s1_22{}},
+		// Ubuntu support
+		{os: "Ubuntu_20.04.1_x86-64", k8s: "v1.22.1", algo: &algo.Ubuntu20_4K8s1_22{}},
```
dharmjit, line 43: Suggestion: Can we have below instead and remove the osFilter and aliasOSMap implementation.
```suggestion
		{os: "Ubuntu_20.04.*_x86-64", k8s: "v1.22.1", algo: &algo.Ubuntu20_4K8s1_22{}},
```

agent/installer/registry.go
```diff
@@ -32,22 +35,49 @@ func (r *registry) Add(os, k8sVer string, installer osk8sInstaller) {
...
+
+	// Try to find os using the alias map
+	for k,v := range r.aliasOSMap {
+		matched, _ := regexp.MatchString(k, os)
+		if matched {
+			return r.osk8sInstallerMap[v]
+		}
+	}
+
+	return nil
```
dharmjit, line 82: Suggestion: If we remove the aliasOSMap implementation, this can be simplified to something like this
```
func (r *registry) resolveK8sInstallerMap(os string) k8sInstallerMap {
	for k := range r.osk8sInstallerMap {
		var re = regexp.MustCompile(k)
		if re.MatchString(os) {
			return r.osk8sInstallerMap[os]
		}
	}
	return nil
}
```

ratanasovvmw: The idea of the map is the same as libraries in linux. In our case the symlink is the osFilter and the bundle is the lib on the right side. The keys of the first map are used for bundle name. Moving them elsewhere will require more refactoring and can be done post FC.

```
lrwxrwxrwx 1 root root      10 юни  4  2020 ld-linux.so.2 -> ld-2.27.so
-rw-r--r-- 1 root root   14024 юни  4  2020 libanl-2.27.so
lrwxrwxrwx 1 root root      14 юни  4  2020 libanl.so.1 -> libanl-2.27.so
-rw-r--r-- 1 root root    5544 юни  4  2020 libBrokenLocale-2.27.so
lrwxrwxrwx 1 root root      23 юни  4  2020 libBrokenLocale.so.1 -> libBrokenLocale-2.27.so
-rwxr-xr-x 1 root root 1926828 юни  4  2020 libc-2.27.so
-rw-r--r-- 1 root root  185920 юни  4  2020 libcidn-2.27.so
lrwxrwxrwx 1 root root      15 юни  4  2020 libcidn.so.1 -> libcidn-2.27.so
-rw-r--r-- 1 root root   38440 юни  4  2020 libcrypt-2.27.so
lrwxrwxrwx 1 root root      16 юни  4  2020 libcrypt.so.1 -> libcrypt-2.27.so

```

agent/installer/installer.go
ratanasovvmw: See below about aliasOSMap. Can be done post FC.

Removing patch version from bundle name will make it *impossible* to do something different for some patch version, It makes too vague what we map to what. What is the benefit of doing so?

dharmjit: @ratanasovvmw This comment takes into consideration my other comment where I thought the implementation can be made simpler by defining the regex for the `os` field itself. But I am ok if we wanted to do this post 0.1.0 release.

agent/installer/registry.go
dharmjit: Thanks for the reference, But do you think there are some advantages to follow the same for our implementation.

ratanasovvmw: Yes, we're essentially doing the same thing.

agent/installer/installer.go
```diff
@@ -33,27 +33,38 @@ type installer struct {
...
 
-	reg := NewRegistry()
-	for _, t := range supportedOsK8s {
-		a := &algo.BaseK8sInstaller{
-			K8sStepProvider: t.algo,
-			/*BundlePath: will be set when tag is known */
-			OutputBuilder: ob}
-		reg.Add(t.os, t.k8s, a)
+		// Match concrete os version to repository os version
+		reg.AddOsFilter("Ubuntu_20.04.*_x86-64", linuxDistro)
```
jamiemonserrate, line 58: Question - do we need to hard code this? Can this be injected/overridden via a config file? This could help in our integration testing too.

agent/installer/cli.go
```diff
@@ -8,11 +8,12 @@ import (
...
 	"github.com/go-logr/logr"
 	"k8s.io/klog/klogr"
+	"os"
+	"text/tabwriter"
 )
 
 var (
-	listSupportedFlag    = flag.Bool("list-supported", false, "List all supported OS and Kubernetes versions")
-	listBundlesFlag      = flag.Bool("list-bundles", false, "List the BYOH Bundle names for all supported OS and Kubernetes versions")
+	listSupportedFlag    = flag.Bool("list-supported", false, "List all supported OS, Kubernetes versions and BYOH Bundle names")
```
yixingjia, line 16: Just a question, maybe I'm wrong :) 
Should we add the list support bundles function in the management cluster side?

If I'm an end user, after I install the byoh provider , then I will expect to check what bundles do we have in the repo on the management cluster side instead of login to the remote byoh host. 

Then I can decide which version of kubernest/ bundle I would like to use to create the cluster on byoh host.

ratanasovvmw: Makes sense. This is just a CLI of the installer lib that is used by the host agent. So, the CP can retrieve the same info by the lib directly. The exact workflow for that is outside installer lib charter.

agent/installer/installer.go
ratanasovvmw: It can be made configurable when needed.

Technically speaking
- only the "Ubuntu_20.04.*_x86-64" can be configurable
- the "Ubuntu_20.04.1_x86-64" cannot be configurable because it is bound to a struct which may be different for OSes and versions. It is the core of the installer

Workflow speaking,
- [yixingjia](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/pull/207#discussion_r743401846) implies that this lib is the source of truth and needs to be queried
- if it is configurable whose responsibility would be to do that?

Testing speaking,
- my experience so far is that we need to override the detection of the current OS. E.g. if lib has ubuntu2004 and I'm running on ubuntu1804, I can override and treat it as ubuntu2004

ratanasovvmw: Had a meeting with Dharmjit and Yixing where config files were also mentioned but regarding that downstream version for k8s is 1.2.1+vmware.1.
Agreed we need to look at this holistically post FC.

agent/installer/cli.go
ratanasovvmw: https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/pull/207#discussion_r743457774

agent/installer/installer.go
dharmjit: Okay for now, but we need to revisit this again.

dharmjit (approved):
Changes look good for now, But as @ratanasovvmw mentioned we need to look at it again from the UX perspective.

PR #216, opened 2021-11-08
Integrate host-agent with the installer

### LINKED ISSUE #204: Integrate host agent with the installer
- host agent should exit if the OS is not supported (i.e., don't wait for host registration)
- call install / uninstall methods of the installer by passing the k8sversion and bundleTag params
- write units tests wherever applicable

anusha94:
Signed-off-by: Anusha Hegde <anushah@vmware.com>

- introduce a new flag skip-installation for the host-agent. The site
  operators can use this flag if they wish to manage the k8s components
  themselves
- refactor installer to update New, Install and Uninstall signatures

Fixes #204


agent/installer/installer.go
```diff
@@ -103,29 +100,28 @@ func New(bundleRepo, downloadPath string, logger logr.Logger) (*installer, error
...
-		logger:           logger}, nil
+		algoRegistry: reg,
+		detectedOs:   currentOs,
+		logger:       logger}, nil
 }
 
 // Install installs the specified k8s version on the current OS
-func (i *installer) Install(k8sVer, tag string) error {
+func (i *installer) Install(bundleRepo, k8sVer, tag string) error {
+	i.bundleDownloader = bundleDownloader{repoAddr: bundleRepo, downloadPath: i.downloadPath, logger: i.logger}
```
ratanasovvmw, line 123: it's bad design that this method modifies the installer object. Same for uninstall.

agent/installer/installer.go
```diff
@@ -88,10 +88,7 @@ func (bd *bundleDownloader) DownloadOrPreview(os, k8s, tag string) error {
 // New returns an installer that downloads bundles for the current OS from OCI repository with
 // address bundleRepo and stores them under downloadPath. Download path is created,
 // if it does not exist.
-func New(bundleRepo, downloadPath string, logger logr.Logger) (*installer, error) {
-	if bundleRepo == "" {
-		return nil, fmt.Errorf("empty bundle repo")
-	}
+func New(downloadPath string, logger logr.Logger) (*installer, error) {
```
ratanasovvmw, line 91: downloadPath goes with bundleRepo. It's a cache of the bundle repo. It's odd that they are decoupled in different methods.

ratanasovvmw (commented):
I recommend to split these reviews in 2 so they are smaller
- integration
- installer refactoring if needed

agent/main.go
```diff
@@ -84,36 +87,39 @@ func main() {
 	flag.StringVar(&namespace, "namespace", "default", "Namespace in the management cluster where you would like to register this host")
 	flag.Var(&labels, "label", "labels to attach to the ByoHost CR in the form labelname=labelVal for e.g. '--label site=apac --label cores=2'")
 	flag.StringVar(&metricsbindaddress, "metricsbindaddress", ":8080", "metricsbindaddress is the TCP address that the controller should bind to for serving prometheus metrics.It can be set to \"0\" to disable the metrics serving")
+	flag.StringVar(&downloadpath, "downloadpath", "/var/lib", "File System path to keep the downloads")
```
yixingjia, line 90: I guess download path is just to store the download files, right? In that way we can just hard-code this folder instead of ask the user to provide this.  Just like other software that have their default install directory.

agent/main.go
```diff
@@ -84,36 +87,39 @@ func main() {
...
+	flag.BoolVar(&skipInstallation, "skip-installation", false, "If you want to skip installation of the kubernetes component binaries")
```
yixingjia, line 91: Kubernetes component binary ?  I guess we only refer to the kubelet / kubeadm and kubectl right?

agent/main.go
ratanasovvmw: This is opt-in, with a default. It should be /var/lib/**someting**

agent/main.go
ratanasovvmw: yes, + containerd, kubernetes-cni and cri-tools

dharmjit (approved):
LGTM!

dharmjit:
We could open a linked Issue in case there are some code improvements or code refactoring is required.

agent/main.go
anusha94: Makes sense. Created a new issue to track this - https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/issues/218

agent/installer/installer.go
jamiemonserrate: Agreed. This needs to be refactored. I think the old assumption was that we would have the `repoAddr` when the host-agent was started. But this assumption was false. The `repoAddr` will be passed in at Install / Uninstall time, so there is no point saving it. We just took the path of least resistance to get the installer integrated.

agent/installer/installer.go
jamiemonserrate: I thought downloadPath is a location on disk to cache the repo. Why would it be coupled to the bundleRepo?

agent/installer/installer.go
jamiemonserrate: Created #221 to track this work

agent/installer/installer.go
ratanasovvmw: Indeed "cache **the** repo" - yes. After this change there is no "the repo"

This change makes the cache shareable between repos. If they happen to have different bundles with the same k8s version, this will simply not work as they will clash and will have false hits. Installer assumes 1 repo, 1 local cache. This change is half-making another model.

agent/installer/installer.go
ratanasovvmw: I acknowledge the least resistance for FC. That was to New every time. A much smaller and safer change.

PR #217, opened 2021-11-08
Strip version tag from k8s version

anusha94:

The byomachine controller will strip any version tag from k8sversion
before setting the annotation on byohost


dharmjit (approved):
LGTM!

PR #243, opened 2021-11-15
Fix installer Ubuntu_20.04.3_x86-64 bug

ratanasovvmw:
Problem
When running cli with os flag set to "Ubuntu_20.04.3_x86-64", installer returns
"os k8s not supported". Installer is expected to support Ubuntu_20.04.*_x86-64
and treat them as Ubuntu_20.04.1_x86-64.

Rootcause
Installer runs a check whether there is at least one k8s for the os.
It does so by using registry ListK8s. However the latter expects that
its argument is an osBundle (Ubuntu_20.04.1_x86-64) as returned by
registry ListOS. There is no os bundle for "Ubuntu_20.04.3_x86-64".

Fix
registry ListK8s needs to support 2 use-cases. First, when called
with an actual OS to be mutated. Second, when called with list-supported
which may be on a non supported OS. Extend registry ListK8s to use the os
regex in addition to the os bundle. Extend registry to return the os bundle
in addition to installer.

### DIFF
--- a/agent/installer/installer.go
+++ b/agent/installer/installer.go
@@ -156,16 +156,18 @@ func (i *installer) Uninstall(k8sVer, tag string) error {
 func (i *installer) getAlgoInstallerWithBundle(k8sVer, tag string) (osk8sInstaller, error) {
 	// This OS supports at least 1 k8s version. See New.
 
-	algoInst := i.algoRegistry.GetInstaller(i.detectedOs, k8sVer)
+	algoInst, osBundle := i.algoRegistry.GetInstaller(i.detectedOs, k8sVer)
 	if algoInst == nil {
 		return nil, ErrOsK8sNotSupported
 	}
+	i.logger.Info("Current OS will be handled as", "OS", osBundle)
+
 	// copy installer from registry and set BundlePath including tag
 	// empty means preview mode
 	algoInstCopy := *algoInst.(*algo.BaseK8sInstaller)
 	algoInstCopy.BundlePath = i.bundleDownloader.getBundlePathDirOrPreview(k8sVer, tag)
 
-	bdErr := i.bundleDownloader.DownloadOrPreview(i.detectedOs, k8sVer, tag)
+	bdErr := i.bundleDownloader.DownloadOrPreview(osBundle, k8sVer, tag)
 	if bdErr != nil {
 		return nil, bdErr
 	}
@@ -198,7 +200,7 @@ func getSupportedRegistryDescription() registry {
 func PreviewChanges(os, k8sVer string) (install, uninstall string, err error) {
 	stepPreviewer := stringPrinter{msgFmt: "# %s"}
 	reg := getSupportedRegistry(&bundleDownloader{}, &stepPreviewer)
-	installer := reg.GetInstaller(os, k8sVer)
+	installer, _ := reg.GetInstaller(os, k8sVer)
 
 	if installer == nil {
 		err = ErrOsK8sNotSupported
--- a/agent/installer/registry.go
+++ b/agent/installer/registry.go
@@ -57,16 +57,30 @@ func (r *registry) ListOS() (osFilter, osBundle []string) {
 	return
 }
 
-func (r *registry) ListK8s(osBundle string) []string {
-	result := make([]string, 0, len(r.osk8sInstallerMap[osBundle]))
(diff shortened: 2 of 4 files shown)


agent/installer/registry.go
```diff
@@ -57,16 +57,30 @@ func (r *registry) ListOS() (osFilter, osBundle []string) {
 	return
 }
 
-func (r *registry) ListK8s(osBundle string) []string {
-	result := make([]string, 0, len(r.osk8sInstallerMap[osBundle]))
-	for k8s := range r.osk8sInstallerMap[osBundle] {
+func (r *registry) ListK8s(osBundleHost string) []string {
+	var result []string
+
+	// os bundle
```
dharmjit, line 63: Suggestion: can we update the comment to something like below.
```
// Check if the hostOS is directly supported
```

agent/installer/registry.go
```diff
@@ -57,16 +57,30 @@ func (r *registry) ListOS() (osFilter, osBundle []string) {
...
+	if k8sMap, ok := r.osk8sInstallerMap[osBundleHost]; ok {
+		for k8s := range k8sMap {
+			result = append(result, k8s)
+		}
+
+		return result
+	}
+
+	// os host
```
dharmjit, line 72: Suggestion: Can we update the comment to something like below.
```
// Check if the hostOS is supported via regex filter
```

agent/installer/registry.go
```diff
@@ -57,16 +57,30 @@ func (r *registry) ListOS() (osFilter, osBundle []string) {
...
+	for k8s := range r.osk8sInstallerMap[r.resolveOsToOsBundle(osBundleHost)] {
 		result = append(result, k8s)
 	}
+
 	return result
 }
 
-func (r *registry) GetInstaller(osHost, k8sVer string) osk8sInstaller {
-	return r.osk8sInstallerMap[r.resolveOsToOsBundle(osHost)][k8sVer]
+func (r *registry) GetInstaller(osHost, k8sVer string) (osk8si osk8sInstaller, osBundle string) {
```
dharmjit, line 80: Suggestion:  Do you think `resolvedOS` sounds better instead of `osBundle` or maybe it has some other meaning.

agent/installer/installer.go
```diff
@@ -156,16 +156,18 @@ func (i *installer) Uninstall(k8sVer, tag string) error {
 func (i *installer) getAlgoInstallerWithBundle(k8sVer, tag string) (osk8sInstaller, error) {
 	// This OS supports at least 1 k8s version. See New.
 
-	algoInst := i.algoRegistry.GetInstaller(i.detectedOs, k8sVer)
+	algoInst, osBundle := i.algoRegistry.GetInstaller(i.detectedOs, k8sVer)
```
dharmjit, line 159:
```suggestion
	algoInst, resolvedOS := i.algoRegistry.GetInstaller(i.detectedOs, k8sVer)
```

dharmjit (commented):
LGTM overall, some nits only.

agent/installer/registry.go
```diff
@@ -57,16 +57,30 @@ func (r *registry) ListOS() (osFilter, osBundle []string) {
 	return
 }
 
-func (r *registry) ListK8s(osBundle string) []string {
-	result := make([]string, 0, len(r.osk8sInstallerMap[osBundle]))
-	for k8s := range r.osk8sInstallerMap[osBundle] {
+func (r *registry) ListK8s(osBundleHost string) []string {
```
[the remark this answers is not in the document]
ratanasovvmw: it's both, it's not only host os.

agent/installer/registry.go
ratanasovvmw: // Check if the hostOS is directly supported
This is one possible usage.

It's just returning the k8s for an os.

agent/installer/registry.go
ratanasovvmw: os bundle - the os as published by a bundle

We're resolving the actual host OS to an os bundle.

dharmjit (approved):
LGTM

anusha94 (approved):
I tested this branch on a Ubuntu 20.04.3 vm and looks good to me.
Used this command - `sudo ./installer-cli -install -bundle-repo <repo_addr> -tag <tag_name> -k8s v1.22.3`

Issue #242, opened 2021-11-15
Only run [PR-Blocking] tests for PRs


jamiemonserrate:
**Describe the solution you'd like**
We should only run the `[PR-Blocking]` tests as part of each PR. This will help improve the feedback time to folks raising PRs. 

We should also have the ability to run the rest of the suite on demand. Preferably via github `/commands`

karuppiah7890:
For K8s Prow like `/command` in GitHub, you can checkout https://github.com/jpmcb/prow-github-actions which is something we use for labelling issues primarily. You could checkout how it can be used to run test suits on GitHub Actions

anusha94:
Thanks @karuppiah7890 for the pointer. @shamsher31 , is this something we can investigate for this project?

shamsher31:
> For K8s Prow like `/command` in GitHub, you could checkout https://github.com/jpmcb/prow-github-actions which is something we use for labelling issues primarily. You could checkout how it can be used to run test suites on GitHub Actions

This looks great. Thanks for sharing @karuppiah7890 .

> @shamsher31 , is this something we can investigate for this project?

We can use it for now until Prow is available across Tanzu projects.

PR #245, opened 2021-11-16
Run e2e test for each PR

shamsher31:
This will run the e2e test for each PR and won't merge unless all the tests are fixed.

Fixes https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/issues/242

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -28,3 +28,26 @@ jobs:
 
     - name: run e2e test
       run: make test-e2e
+  
+  e2e-pr-blocking:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
+
+    - name: Set up Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1.16
+
+    - name: Install ginkgo
+      run: sudo apt-get install golang-ginkgo-dev
+
+    - name: turn off swap
+      run: sudo swapoff -a
+
+    - name: Set netfilter conntrack max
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -21,27 +21,3 @@ jobs:
 
     - name: Run test make target
       run: make test
-
-
-  e2e-pr-blocking:
```
anusha94, line 26: Why are we removing this? I think it is good to run quick start test on merge to main. WDYT?

.github/workflows/e2e.yml
```diff
@@ -28,3 +28,26 @@ jobs:
 
     - name: run e2e test
       run: make test-e2e
+  
+  e2e-pr-blocking:
```
anusha94, line 32: this should replace the above `make test-e2e` which runs all e2e tests. This change is running all e2e tests + additional run of quick start

.github/workflows/main.yml
shamsher31: As the name suggests `e2e-pr-blocking` should only run before PR is merged. Once all the checks are passed and PR is approved and merged, it will be a waste of resources and time to run those tests again?

.github/workflows/e2e.yml
shamsher31: Yup, make sense.

.github/workflows/main.yml
anusha94: Yup. Makes sense to me. Okay to remove 👍🏽

anusha94 (approved):
lgtm

dharmjit (approved):
LGTM!

PR #255, opened 2021-11-18
Installer e2e test

ratanasovvmw:
This test executes the following workflow:

- download upstream bundle contents
- start local repo
- build a BYOH bundle and upload it to the local repo
- spin up a Ubuntu 20.04 VM and inside it run
  - cli install with local repo
  - kubeadm init phase preflight
  - cli uninstall

Upon success return 0.

### DIFF
--- a/docs/local_dev.md
+++ b/docs/local_dev.md
@@ -303,7 +303,8 @@ This test executes the following workflow:
 - start local repo
 - build a BYOH bundle and upload it to the local repo
 - spin up a Ubuntu 20.04 VM and inside it run
--- cli install with local repo
--- kubeadm init phase preflight
--- cli uninstall
+  - cli install with local repo
+  - kubeadm init phase preflight
+  - cli uninstall
+
 Upon success return 0.


jamiemonserrate:
@ratanasovvmw - This e2e test seems to test the installer in isolation - so its _not really_ an e2e test then :) 

I think this should be integrated within our existing e2e suite. More specifically, when we run the [PR-Blocking](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/main/test/e2e/e2e_test.go#L31) test, the installer should be triggered as part of that workflow. 

So what that means is 
- That test would need to build the bundle and host it (So [these bash lines](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/0301cd4d06811bddf6653756454e7def1ac298bd/agent/installer/e2e/test.sh#L29-L36) move into the golang test itself - probably [e2e_suite_test.go](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/main/test/e2e/e2e_suite_test.go))
- We would need to start Vagrant in our [e2e_suite_test.go](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/main/test/e2e/e2e_suite_test.go)
- The e2e tests should then spin up all containers inside this Vagrant VM. So that would mean modifying the [e2e_docker_helper.go](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/main/test/e2e/e2e_docker_helper.go#L134) to do this. 
- We would then stop invoking `--skip-installation` - as we do [here](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/main/test/e2e/e2e_docker_helper.go#L194)

That is a proper e2e test in my mind. Thoughts?

jamiemonserrate (changes requested):
@ratanasovvmw - This e2e test seems to test the installer in isolation - so its _not really_ an e2e test then :) 

I think this should be integrated within our existing e2e suite. More specifically, when we run the [PR-Blocking](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/main/test/e2e/e2e_test.go#L31) test, the installer should be triggered as part of that workflow. 

So what that means is 
- That test would need to build the bundle and host it (So [these bash lines](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/0301cd4d06811bddf6653756454e7def1ac298bd/agent/installer/e2e/test.sh#L29-L36) move into the golang test itself - probably [e2e_suite_test.go](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/main/test/e2e/e2e_suite_test.go))
- We would need to start Vagrant in our [e2e_suite_test.go](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/main/test/e2e/e2e_suite_test.go)
- The e2e tests should then spin up all containers inside this Vagrant VM. So that would mean modifying the [e2e_docker_helper.go](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/main/test/e2e/e2e_docker_helper.go#L134) to do this. 
- We would then stop invoking `--skip-installation` - as we do [here](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/main/test/e2e/e2e_docker_helper.go#L194)

That is a proper e2e test in my mind. Thoughts?

ratanasovvmw:
@jamiemonserrate
My idea was exactly to make an installer e2e test :) (currently the test is under agent/installer/e2e) as an overview how it will look like. Whether it'll go in like this, or be integrated in the project e2e tests is another thing. In general, I think there should be both installer e2e and project e2e so failures are localized faster.

To spin up a container in Vagrant (initially I was going to do so) you need containerd etc and configure it in the Vagrant VM. This is what installer is doing. So, imagine that part of installer is removed. The test will still pass because they are already there.
There might be other ways to get around this.

PR #257, opened 2021-11-19
Update Cluster Template for ByoCluster Bundle fields

### LINKED ISSUE #252: Add bundleLookupBaseRegistry, bundleLookupTag in cluster-template for ByoCluster
With this #205 PR, we introduced 2 new fields `bundleLookupBaseRegistry`, `bundleLookupTag` for `ByoCluster` CRD. We need to define the same in `cluster-template.yaml` with their default /useful values.

We need to
- Add these fields to  [cluster-with-kcp.yaml](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/main/test/e2e/data/infrastructure-provider-byoh/v1beta1/cluster-with-kcp.yaml#L25).
- Run `make cluster-templates` to update the cluster-template.yaml.

dharmjit:
Added bundleLookupBaseRegistry,bundleLookupTag for ByoCluster in cluster templates

This PR will update the cluster template to align with the latest ByoCluster Spec

Fixes #252 

**Additional information**

### DIFF
--- a/test/e2e/data/infrastructure-provider-byoh/v1beta1/cluster-template.yaml
+++ b/test/e2e/data/infrastructure-provider-byoh/v1beta1/cluster-template.yaml
@@ -183,6 +183,8 @@ kind: ByoCluster
 metadata:
   name: ${CLUSTER_NAME}
 spec:
+  bundleLookupBaseRegistry: projects.registry.vmware.com/cluster_api_provider_bringyourownhost
+  bundleLookupTag: v0.1.0_alpha.2
   controlPlaneEndpoint:
     host: ${CONTROL_PLANE_ENDPOINT_IP}
     port: 6443
--- a/test/e2e/data/infrastructure-provider-byoh/v1beta1/cluster-with-kcp.yaml
+++ b/test/e2e/data/infrastructure-provider-byoh/v1beta1/cluster-with-kcp.yaml
@@ -27,6 +27,8 @@ kind: ByoCluster
 metadata:
   name: ${CLUSTER_NAME}
 spec:
+  bundleLookupBaseRegistry: projects.registry.vmware.com/cluster_api_provider_bringyourownhost
+  bundleLookupTag: v0.1.0_alpha.2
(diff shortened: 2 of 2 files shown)


test/e2e/data/infrastructure-provider-byoh/v1beta1/cluster-template.yaml
```diff
@@ -183,6 +183,8 @@ kind: ByoCluster
 metadata:
   name: ${CLUSTER_NAME}
 spec:
+  bundleLookupBaseRegistry: projects.registry.vmware.com/cluster_api_provider_bringyourownhost
+  bundleLookupTag: v0.1.0_alpha.2
```
shamsher31, line 187: Is this going to be updated with every release, if yes then I think its better to use it from variable?

dharmjit: @shamsher31 It is not going to change frequently for a bundle of specific k8s version/OS, But I agree with what you suggested. Let others provide their opinion as well.

anusha94: I'm okay with this for short-term. But for longer term, IMO, this field should be optional and if not provided should be treated as `latest` value. Not very sure how tagging works. Does this mean we will have 2 tags for the same image? Is that even allowed?

Otherwise, makes sense to read from a variable.
Do we open a new issue for this and discuss options there? This PR is good to go! (after we have the artifact in the registry, of course!)

lwmqwer (approved):
Let merge now, as we have issue to track it, I think the webhook for deafult it the final solution

anusha94 (approved):

test/e2e/data/infrastructure-provider-byoh/v1beta1/cluster-template.yaml
dharmjit: Created this Issue #259 to decide on the other options.

PR #287, opened 2021-12-14
Define strategy to provide value for bundleLookupTag in Cluster Template

### LINKED ISSUE #259: Define strategy to provide value for bundleLookupTag in Cluster Template
Derived from the discussion [here](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/pull/257#discussion_r752910834), we need to decide on a strategy to provide the value for `bundleLookupTag`. It could be 
- Variable based
- Defaulting Webhook
- `latest` image tag

huchen2021:
Fixes #259 

Signed-off-by: Hui Chen <huchen@huchen-a01.vmware.com>


apis/infrastructure/v1beta1/byocluster_webhook.go
```diff
@@ -0,0 +1,65 @@
...
+
+var _ webhook.Defaulter = &ByoCluster{}
+
+// Default implements webhook.Defaulter so a webhook will be registered for the type
+// nolint: stylecheck
+func (r *ByoCluster) Default() {
+	byoclusterlog.Info("default", "name", r.Name)
+
+	if r.Spec.BundleLookupTag == "" {
+		r.Spec.BundleLookupTag = "v0.1.0_alpha.2"
```
lwmqwer, line 34: we'd better define an const for this and collect them together in a file and generate from current release

anusha94: +1 for reading from const

> collect them together in a file and generate from current release

This is still adding complexity IMO, we should be good with reading from an env variable while applying the cluster-template. Default value should be something like `latest` - But in our case, this default value may not be applicable yet

We can make this field **required** for now, it will error out if the field is not set. Unless we figure out a better way to publish bundles with `latest` tag to the repository, I suggest we leave it as is.

dharmjit: I am aligned with @anusha94 suggestion as we haven't yet figured to tag appropriately the various bundles. Reading it for an env variable seems like a good option to me as of now.

Please note that currently we only have a single bundle for k8s 1.22.3 and ubuntu 20.04 but going forward we could have various bundles which might can increase the complexity of naming/tagging bundles.

huchen2021: @anusha94 I accepted your comment. Please review it again. @dharmjit

apis/infrastructure/v1beta1/byocluster_webhook.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+// TODO(user): change verbs to "verbs=create;update;delete" if you want to enable deletion validation.
+//+kubebuilder:webhook:path=/validate-infrastructure-cluster-x-k8s-io-v1beta1-byocluster,mutating=false,failurePolicy=fail,sideEffects=None,groups=infrastructure.cluster.x-k8s.io,resources=byoclusters,verbs=create;update,versions=v1beta1,name=vbyocluster.kb.io,admissionReviewVersions=v1
+
+var _ webhook.Validator = &ByoCluster{}
+
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
+func (r *ByoCluster) ValidateCreate() error {
+	byoclusterlog.Info("validate create", "name", r.Name)
```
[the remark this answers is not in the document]
huchen2021: Yes. we need this.

apis/infrastructure/v1beta1/byocluster_webhook.go
```diff
@@ -0,0 +1,73 @@
...
+	if r.Spec.BundleLookupTag == "" {
+		return apierrors.NewForbidden(groupResource, r.Name, errors.New("cannot create ByoCluster with Spec.BundleLookupTag"))
+	}
+
+	return nil
+}
+
+// ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
+func (r *ByoCluster) ValidateUpdate(old runtime.Object) error {
+	byoclusterlog.Info("validate update", "name", r.Name)
```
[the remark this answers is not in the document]
huchen2021: Yes. we need this.

apis/infrastructure/v1beta1/byocluster_webhook.go
```diff
@@ -0,0 +1,73 @@
...
+
+	if r.Spec.BundleLookupTag == "" {
+		return apierrors.NewForbidden(groupResource, r.Name, errors.New("cannot create ByoCluster with empty Spec.BundleLookupTag"))
+	}
+	return nil
+}
+
+// ValidateDelete implements webhook.Validator so a webhook will be registered for the type
+func (r *ByoCluster) ValidateDelete() error {
+	byoclusterlog.Info("validate delete", "name", r.Name)
```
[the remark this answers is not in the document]
huchen2021: ok.

PROJECT
```diff
@@ -32,6 +32,10 @@ resources:
   kind: ByoCluster
   path: github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/apis/infrastructure/v1beta1
   version: v1beta1
+  webhooks:
+    defaulting: true
```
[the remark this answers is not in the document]
huchen2021: Keep this for future. It would be save our energy if we want add something for Default.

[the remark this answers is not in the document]
anusha94: I'm also actually +1 for removing this. Let's add it only when needed. Keeping only things that we need now will reduce maintenance overhead.

apis/infrastructure/v1beta1/byocluster_webhook.go
```diff
@@ -0,0 +1,68 @@
...
+
+var _ webhook.Validator = &ByoCluster{}
+
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
+func (r *ByoCluster) ValidateCreate() error {
+	byoclusterlog.Info("validate create", "name", r.Name)
+	groupResource := schema.GroupResource{Group: "infrastructure.cluster.x-k8s.io", Resource: "byocluster"}
+
+	if r.Spec.BundleLookupTag == "" {
+		return apierrors.NewForbidden(groupResource, r.Name, errors.New("cannot create ByoCluster without Spec.BundleLookupTag"))
```
anusha94, line 46: Does [func NewInvalid](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/errors#NewInvalid) make more sense here?
Thoughts? @huchen2021 @dharmjit

apis/infrastructure/v1beta1/byocluster_webhook.go
```diff
@@ -0,0 +1,68 @@
...
+	groupResource := schema.GroupResource{Group: "infrastructure.cluster.x-k8s.io", Resource: "byocluster"}
+
+	if r.Spec.BundleLookupTag == "" {
+		return apierrors.NewForbidden(groupResource, r.Name, errors.New("cannot update ByoCluster with empty Spec.BundleLookupTag"))
+	}
+	return nil
+}
+
+// ValidateDelete implements webhook.Validator so a webhook will be registered for the type
+func (r *ByoCluster) ValidateDelete() error {
```
anusha94, line 65: Good to remove this `func` to keep the file more clean.

apis/infrastructure/v1beta1/byocluster_webhook_test.go
```diff
@@ -0,0 +1,87 @@
...
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/kubectl/pkg/scheme"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var _ = Describe("ByoclusterWebhook", func() {
+	XContext("When ByoCluster gets a create request", func() {
```
anusha94, line 16:
```suggestion
	Context("When ByoCluster gets a create request", func() {
```

Does this test fail? Why are we ignoring it?

apis/infrastructure/v1beta1/byocluster_webhook_test.go
```diff
@@ -0,0 +1,87 @@
...
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "byocluster-create",
+					Namespace: "default",
+				},
+				Spec: ByoClusterSpec{},
+			}
+		})
+
+		It("should reject the request when BundleLookupTag is empty", func() {
+			err := byoCluster.ValidateCreate()
```
anusha94, line 38: An integration test is more useful here - using testEnv etc. like we do for controllers.

apis/infrastructure/v1beta1/byocluster_webhook_test.go
huchen2021: The test failed when I use Context function.

apis/infrastructure/v1beta1/byocluster_webhook.go
dharmjit: No strong views, But looking at the `cluster-api` project code,`NewInvalid` is mostly used and came to know that NewInvalid can provide more context in the error - [slack conversation ](https://kubernetes.slack.com/archives/C8TSNPY4T/p1575574280347600)

apis/infrastructure/v1beta1/byocluster_webhook_test.go
huchen2021: This is not integration test , I refer it [byohost_webhook_test](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/main/apis/infrastructure/v1beta1/byohost_webhook_test.go)

apis/infrastructure/v1beta1/byocluster_webhook.go
huchen2021: This function is auto-generated, no need to remove it. Actually, this file is auto-generated. It's better to fill what you need, not delete what you don't need.

apis/infrastructure/v1beta1/byocluster_webhook.go
huchen2021: ok

PROJECT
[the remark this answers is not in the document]
huchen2021: But if we want to  auto-generate the default function again, we need to delete the whole ByoCluster webhook first. Or it will report the following error:
```
2022/01/11 02:44:11 failed to create webhook: unable to inject the resource to "base.go.kubebuilder.io/v3": webhook resource already exists
```
This will cause extra burden。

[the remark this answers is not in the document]
anusha94: At that time, we only write this new function. It doesn't necessarily have to be generated via kubebuilder. I'm stressing on removing this because this is the code without any tests and no usage. So we shouldn't have this. Makes sense? 

Probably we can keep this entry in the `PROJECT` file and remove the unused functions from the byocluster_webhook file. What do you think?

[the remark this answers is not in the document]
huchen2021: @anusha94 ok.

apis/infrastructure/v1beta1/byocluster_webhook_test.go
huchen2021: Fix it.

apis/infrastructure/v1beta1/byocluster_webhook_test.go
huchen2021: FIx it.

PROJECT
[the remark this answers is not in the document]
anusha94: You will have to delete just these
```
var _ webhook.Defaulter = &ByoCluster{}

// Default implements webhook.Defaulter so a webhook will be registered for the type
// nolint: stylecheck
func (r *ByoCluster) Default() {
}
```
Okay fine let's keep it. No harm I guess.

apis/infrastructure/v1beta1/byocluster_webhook_test.go
```diff
@@ -0,0 +1,88 @@
...
+				},
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "byocluster-create",
+					Namespace: "default",
+				},
+				Spec: ByoClusterSpec{},
+			}
+		})
+
+		It("should reject the request when BundleLookupTag is empty", func() {
```
anusha94, line 43: I think we can test for 2 cases here.
1. when `BundleLookupTag` is not present
2. when `BundleLookupTag` is empty

It should reject request in both cases.

anusha94: Also another case when there is no error - positive scenario.

apis/infrastructure/v1beta1/byocluster_webhook_test.go
```diff
@@ -0,0 +1,88 @@
...
+			err := k8sClientUncached.Create(ctx, byoCluster)
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring("cannot create ByoCluster without Spec.BundleLookupTag"))
+		})
+
+	})
+
+	Context("When ByoCluster gets a update request", func() {
```
anusha94, line 51:
```suggestion
	Context("When ByoCluster gets an update request", func() {
```

apis/infrastructure/v1beta1/byocluster_webhook_test.go
```diff
@@ -0,0 +1,88 @@
...
+					Namespace: "default",
+				},
+				Spec: ByoClusterSpec{},
+			}
+		})
+
+		It("should reject the request when BundleLookupTag is empty", func() {
+			err := k8sClientUncached.Create(ctx, byoCluster)
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring("cannot create ByoCluster without Spec.BundleLookupTag"))
```
anusha94, line 46: Let's match full error if possible. Like this -
https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/2a1aaaa4cae69640fdd71f7f1271961543e9fcd2/apis/infrastructure/v1beta1/byohost_webhook_test.go#L85

anusha94 (commented):
minor nits.

PROJECT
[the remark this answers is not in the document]
huchen2021: Good, thanks.

apis/infrastructure/v1beta1/byocluster_webhook_test.go
huchen2021: ok

apis/infrastructure/v1beta1/byocluster_webhook_test.go
huchen2021: This second 

> I think we can test for 2 cases here.
> 
>     1. when `BundleLookupTag` is not present
> 
>     2. when `BundleLookupTag` is empty
> 
> 
> It should reject request in both cases.

For Go-programming, actually there is no such "one field not present". All field are all present, it just initialize or not.

anusha94: I see, okay. Sounds good then. Thanks!

huchen2021: > Also another case when there is no error - positive scenario.

Ok.

apis/infrastructure/v1beta1/byocluster_webhook_test.go
```diff
@@ -0,0 +1,104 @@
...
+		})
+
+		It("should reject the request when BundleLookupTag is empty", func() {
+			byoCluster.Spec.BundleLookupTag = ""
+			err := k8sClientUncached.Update(ctx, byoCluster)
+			Expect(err).To(HaveOccurred())
+			Expect(err).To(MatchError("admission webhook \"vbyocluster.kb.io\" denied the request: ByoCluster.infrastructure.cluster.x-k8s.io \"" + byoCluster.Name + "\" is invalid: <nil>: Internal error: cannot update ByoCluster with empty Spec.BundleLookupTag"))
+		})
+
+		It("should success when BundleLookupTag is not empty", func() {
```
anusha94, line 97:
```suggestion
		It("should update the cluster BundleLookupTag is not empty", func() {
```

apis/infrastructure/v1beta1/byocluster_webhook_test.go
```diff
@@ -0,0 +1,104 @@
...
+			byoCluster.Spec.BundleLookupTag = "fake"
+			err := k8sClientUncached.Update(ctx, byoCluster)
+			Expect(err).NotTo(HaveOccurred())
```
anusha94, line 100: How about getting the cluster object and then checking if value got updated?

huchen2021: ok

anusha94 (approved):
Thank you for the PR!
lgtm

dharmjit (approved):
lgtm! Thank you

PR #299, opened 2022-01-10
Fix integration test for byohost webhook

### LINKED ISSUE #297: Integration test for ByoHost webhook
An attempt was made earlier to write integration test for byohost webhook (which is more useful in checking byohost deletion under different circumstances). Here is the code snippet -
https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/970c5de3c5100f8fb93b6ac85516e6ef735261ec/apis/infrastructure/v1beta1/byohost_webhook_test.go#L19-L21

We are currently ignoring this test by marking it as `XContext`. Debug the issue and enable the test and do any modifications as necessary.

anusha94:
The scheme was not properly registered. So when trying to run the test via testEnv ended up with empty GVK error.
Now that we have integration test, the unit test can be removed.

Fixes #297

### DIFF
--- a/apis/infrastructure/v1beta1/byohost_webhook_test.go
+++ b/apis/infrastructure/v1beta1/byohost_webhook_test.go
@@ -16,9 +16,8 @@ import (
 )
 
 var _ = Describe("ByohostWebhook", func() {
-	// TODO: Make the below Context work
-	// we will not need the second context if this works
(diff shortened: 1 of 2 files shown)


dharmjit (approved):
LGTM

huchen2021 (approved):

PR #308, opened 2022-01-11
Docker cluster template

### LINKED ISSUE #277: Publish different cluster-template artifacts for Docker and everything else
Our e2e tests use Docker container as hosts and the cluster-template used in e2e tests can be found [here](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/main/test/e2e/data/infrastructure-provider-byoh/v1beta1/cluster-template.yaml). However, while publishing the release artifacts, we are manipulating this file to remove Docker specific cgroup sections. This leads to confusion for a user who wants to just download the cluster-template and work with infra of their choice.

We can instead publish 2 cluster-templates - one for Docker based hosts and the other for VM based hosts.

anusha94:
Creating separate cluster templates in the release artifacts - one for docker based byohosts and the other for vm based byohosts
e2e cluster templates are in a separate e2e directory - no release artifacts for this

Test release on my forked repo - https://github.com/anusha94/cluster-api-provider-bringyourownhost/releases/tag/v0.0.7
Diff:
```
kokoni at kangaroo in ~ ah
$ diff cluster-template.yaml new-artifacts/cluster-template.yaml
156c156
<   bundleLookupTag: v0.1.0_alpha.2
---
>   bundleLookupTag: ${BUNDLE_LOOKUP_TAG}

kokoni at kangaroo in ~ ah
$ diff cluster-template-docker.yaml new-artifacts/cluster-template-docker.yaml
18,20d17
<   labels:
<     cni: ${CLUSTER_NAME}-crs-0
<     crs: "true"
167c164
<   bundleLookupTag: v0.1.0_alpha.2
---
>   bundleLookupTag: ${BUNDLE_LOOKUP_TAG}
```

Fixes #277


jamiemonserrate (commented):
Could kustomize help avoid doing everything in `sed`? We seem to use it elsewhere. Any reason we not using it here?

anusha94:
>Could kustomize help avoid doing everything in sed?

This did cross my mind. But looks like kustomize edit can do only so many things
```
$ kustomize edit --help
Edits a kustomization file

Usage:
  kustomize edit [command]

Available Commands:
  add                       Adds an item to the kustomization file.
  alpha-list-builtin-plugin [Alpha] List the builtin plugins
  fix                       Fix the missing fields in kustomization file
  remove                    Removes items from the kustomization file.
  set                       Sets the value of different fields in kustomization file.
```

`remove` is to remove items from kustomization file and not from the files referenced there.

Even if we consider setting fields to empty instead of removing them, `kustomize edit set` is limited to a few fields.
```
$ kustomize edit set --help
Sets the value of different fields in kustomization file.

Usage:
  kustomize edit set [command]

Available Commands:
  annotation  Sets one or more commonAnnotations in kustomization.yaml
  image       Sets images and their new names, new tags or digests in the kustomization file
  label       Sets one or more commonLabels in kustomization.yaml
  nameprefix  Sets the value of the namePrefix field in the kustomization file.
  namespace   Sets the value of the namespace field in the kustomization file
  namesuffix  Sets the value of the nameSuffix field in the kustomization file.
  replicas    Sets replicas count for resources in the kustomization file
```


But I do hear you. Doing so many sed & perl commands on a release artifact may not always be reliable. How about we maintain 2 sets of cluster templates in the repo itself?
The current one is for docker [here](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/tree/main/test/e2e/data/infrastructure-provider-byoh/v1beta1). We can have another folder containing vm related templates. Only downside being if there's any change in the API or defaults, we need to update in 2 places. Hoping these won't be as frequent, this could be a good trade-off I guess.

Thoughts? @jamiemonserrate @dharmjit

dharmjit:
> Hoping these won't be as frequent, this could be a good trade-off I guess

Having explicit templates for VM/docker would be easier to understand the generation of different flavors of `cluster-template` release artifacts. I agree it will add some more maintenance but should be okay I guess.

jamiemonserrate:
Agree with where this thread is heading. 

I can see that the AWS provider is doing something similar - that is, they have a [kustomize_sources](https://github.com/kubernetes-sigs/cluster-api-provider-aws/tree/main/test/e2e/data/infrastructure-aws/kustomize_sources) folder, just like we are planning to have. 

So that reassures me that we are heading in the right direction 😄

Makefile
```diff
@@ -178,8 +179,10 @@ $(RELEASE_DIR):
 build-release-artifacts: build-cluster-templates build-infra-yaml build-metadata-yaml build-host-agent-binary
 
 build-cluster-templates: $(RELEASE_DIR) cluster-templates
-	cp $(BYOH_TEMPLATES)/v1beta1/cluster-template.yaml $(RELEASE_DIR)/cluster-template.yaml
-	sed -i -e 1,20d $(RELEASE_DIR)/cluster-template.yaml
+	cp $(BYOH_TEMPLATES)/v1beta1/templates/docker/cluster-template.yaml $(RELEASE_DIR)/cluster-template-docker.yaml
+	sed -i -e 1,20d $(RELEASE_DIR)/cluster-template-docker.yaml
```
jamiemonserrate, line 183: We paired, and debated that we can avoid doing any `sed` commands by having an additional folder, to separate out the `e2e` and `docker` use cases.

dharmjit: This is there to remove resources `ClusterResourceSet` and associated `ConfigMap` in the generated templates, which are not currently being used.

dharmjit (dismissed):
lgtm!

Makefile
anusha94: We could use `kustomize edit remove resource`. But we think it is better to have separate sources. Created e2e directory for e2e tests - this will not be published during release. The release artifacts are from `docker` and `vm` directories **only**

Makefile
```diff
@@ -182,8 +186,9 @@ $(RELEASE_DIR):
 build-release-artifacts: build-cluster-templates build-infra-yaml build-metadata-yaml build-host-agent-binary
 
 build-cluster-templates: $(RELEASE_DIR) cluster-templates
-	cp $(BYOH_TEMPLATES)/v1beta1/cluster-template.yaml $(RELEASE_DIR)/cluster-template.yaml
-	sed -i -e 1,20d $(RELEASE_DIR)/cluster-template.yaml
```
dharmjit, line 186: Question: Will keeping the `CRS/CM` in the cluster template is going to have any impact? Do we need to test it?

anusha94: We are not keeping it in the vm and docker templates.

dharmjit (approved):

jamiemonserrate (approved):

PR #313, opened 2022-01-13
e2e test with installer

anusha94:
- removed the `skip-installation` flag from e2e test
- updated `test/e2e/BYOHDockerfile` to **exclude** k8s components
- added a few additional required packages (`ufw` and `dbus`) to the Dockerfile
- added a new dev-dockerfile to continue having the same experience in
  getting started and local-dev guides

Successful e2e run: https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/runs/4800081237?check_suite_focus=true

**Notes for Reviewer:**
Adding additional Dockerfile so that the current getting started guides are not affected by this change. This PR's aim is to enable running e2e tests in GH Actions without skipping the installer workflow. Currently, the getting started guides skip this step and we want to continue doing that until we figure out a better way to not modify host config or some other alternative.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,7 @@ RELEASE_DIR := _dist
 # Image URL to use all building/pushing image targets
 IMG ?= ${STAGING_REGISTRY}/${IMAGE_NAME}:${TAG}
 BYOH_BASE_IMG = byoh/node:v1.22.3
+BYOH_BASE_IMG_DEV = byoh-dev/node:v1.22.3
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
 CRD_OPTIONS ?= "crd:trivialVersions=true,preserveUnknownFields=false"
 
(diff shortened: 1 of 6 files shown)


anusha94:
@jamiemonserrate In the e2e test, we [don't have to start the agent with sudo](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/pull/313/files#diff-c31cea32231a99b00dee3330b8e02fbf6286062809855dde7610b6814c77318dR194) because the container itself it started with `privileged = true`
https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/26b4bc2c63fcfb44504ef1817efc2fcf3789ec94/test/e2e/e2e_docker_helper.go#L141

@jamiemonserrate @dharmjit 
wdyt of the contents of `BYOHDockerfile`? Currently I've removed `containerd` and other `k8s` bits. Anything else that the `installer` is taking care of and shouldn't be present in the Dockerfile?

jamiemonserrate (approved):
Reviewed. This LGTM. 

We still have the issue that the tests will change kernel settings for users who attempt to run these tests locally. But lets fix that in a subsequent PR.

dharmjit (approved):
This looks good to me.

Nitpick only, Is there some other dir to keep `BYOHDockerFileDev`. Do you think`BYOHDevDockerFile` is a better name.

anusha94:
> This looks good to me.
> 
> Nitpick only, Is there some other dir to keep `BYOHDockerFileDev`. Do you think`BYOHDevDockerFile` is a better name.

Yea I can change the filename. Where do you think we can keep this file? I kept in the `docs` folder as it is being used there

dharmjit:
> Where do you think we can keep this file?

My initial thoughts were `docs` should only contain documentation and not code. Do you think root dir could be an alternate?

anusha94:
> My initial thoughts were `docs` should only contain documentation and not code. Do you think root dir could be an alternate?

Will do this separately as part of a different PR. Merged this so that other PRs can benefit from running a full e2e test.

Issue #221, opened 2021-11-08
Refactor the Install/Uninstall methods of the installer


jamiemonserrate:
**Describe the solution you'd like**
The Install/Uninstall methods currently modify the state of the installer - which is a bad design. See discussion [here](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/pull/216#discussion_r744643054)

This code was written when we assumed that we would have the bundleRepo Address at startup time. This assumption is not true, and we will only know this information at Install/Uninstall time. 

So let's refactor the installer to reflect this updated assumption.

jeuxdemains:
Me and Alex will take this one.

jeuxdemains:
Here is an update on the task after gathering info about what we need to do:

- during instantiation of the Installer, it should not require or depend on the Bundle Registry which is currently passed to the installer.New() method at creation time (see New() in installer.go, the argument is called bundleRepo)
- once instantiated, the Installer should receive the Bundle Registry (bundleRepo) via setter so it could be set and/or updated anytime
 
The final goal is to have the Installer instance created just once during the lifetime of the installation process and manage its Bundle Registry on-the-fly.
This will also enable performing basic checks (supported OS, pre-requisite packages integrity, etc.) at earlier stage of the installation procedure, at point when the Bundle Registry is not yet known.

jamiemonserrate:
> once instantiated, the Installer should receive the Bundle Registry (bundleRepo) via setter so it could be set and/or updated anytime

I don't think this should be via a setter. I think this should be part of the `Install` contract itself. Otherwise the caller will have to remember to call the setter before calling `Install`, which is confusing and error prone.

jeuxdemains:
You're right. As part of the contract between the user and the Installer is a better solution. Will be updated. 

Download path will be updated as well to avoid clashes.

PR #316, opened 2022-01-13
 Refactor of the host agent installer

georgievaVMW:
In response to this issue: https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/issues/221

Made changes to the methods New and newUnchecked (removed bundleRepo argument)
Removed BundleDownloader argument from getSupportedRegistry as it was not used
Made changes to installer tests and cli to work with the changes.
Made changes to agent/reconciler/host_reconciler.go so tests run

### DIFF
--- a/agent/installer/bundle_downloader_test.go
+++ b/agent/installer/bundle_downloader_test.go
@@ -84,6 +84,25 @@ var _ = Describe("Byohost Installer Tests", func() {
 				mi.Get)
 			Expect(err).ShouldNot(HaveOccurred())
 		})
+		It("Should create and rename dir correctly after successful download", func() {
+			bd.repoAddr = "repo.ccoomm/r/"
+			bd.downloadPath = filepath.Join(bd.downloadPath)
+			err := bd.DownloadFromRepo(
+				normalizedOsVersion,
+				k8sVersion,
+				testTag,
+				mi.Get)
+			Expect(err).ShouldNot((HaveOccurred()))
+			_, err = os.Stat(bd.GetBundleDirPath(k8sVersion, testTag))
+			Expect(err).ShouldNot((HaveOccurred()))
+			notExist := os.IsNotExist(err)
+			Expect(notExist).ShouldNot(BeTrue())
+
+			_, err = os.Stat(bd.GetBundleDirPath(k8sVersion+"a", testTag))
+			Expect(err).Should((HaveOccurred()))
+			notExist = os.IsNotExist(err)
+			Expect(notExist).Should(BeTrue())
+		})
 	})
 	Context("When there is error during download", func() {
 		It("Should return error if given bad repo", func() {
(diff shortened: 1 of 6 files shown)


agent/reconciler/host_reconciler.go
```diff
@@ -289,10 +290,11 @@ func (r *HostReconciler) installK8sComponents(ctx context.Context, byoHost *infr
 	bundleRegistry := byoHost.GetAnnotations()[infrastructurev1beta1.BundleLookupBaseRegistryAnnotation]
 	k8sVersion := byoHost.GetAnnotations()[infrastructurev1beta1.K8sVersionAnnotation]
 	byohBundleTag := byoHost.GetAnnotations()[infrastructurev1beta1.BundleLookupTagAnnotation]
-	bundleInstaller, err := installer.New(bundleRegistry, r.DownloadPath, logger)
+	bundleInstaller, err := installer.New(r.DownloadPath, logger)
 	if err != nil {
 		return err
 	}
+	bundleInstaller.SetBundleRepo(bundleRegistry)
```
jamiemonserrate, line 297: Can we move this `bundleRegistry` to be part of the `Install` contract?

Why? - Because in the current implementation, we are relying on the caller to remember to call `SetBundleRepo` before calling the `Install` method. Lets just make it part of the `Install` method, so that the caller knows that this is the information needed before installing.

jamiemonserrate (dismissed):
My only nit is that maybe we can add a unit test, that asserts we are download to a new path? Other than that, the change LGTM.

agent/installer/bundle_downloader_test.go
```diff
@@ -84,6 +84,25 @@ var _ = Describe("Byohost Installer Tests", func() {
...
+		It("Should create and rename dir correctly after successful download", func() {
+			bd.repoAddr = "repo.ccoomm/r/"
+			bd.downloadPath = filepath.Join(bd.downloadPath)
+			err := bd.DownloadFromRepo(
+				normalizedOsVersion,
+				k8sVersion,
+				testTag,
+				mi.Get)
+			Expect(err).ShouldNot((HaveOccurred()))
+			_, err = os.Stat(bd.GetBundleDirPath(k8sVersion, testTag))
```
jamiemonserrate, line 96: **Nit** - We should hard code the expected values in the test (instead of calling `GetBundleDirPath`. This will
- Improve readability. Because now readers will know what to expect
- Improve test resilience. If there is a bug in `GetBundleDirPath`, it won't affect a huge number of tests.

agent/installer/installer_test.go
```diff
@@ -38,11 +32,11 @@ var _ = Describe("Byohost Installer Tests", func() {
 			for _, os := range osList {
 				i := NewPreviewInstaller(os, nil)
 
-				err := i.Install("unsupported-k8s", testTag)
-				Expect(err).Should(HaveOccurred())
+				err := i.Install("", "unsupported-k8s", testTag)
+				Expect(err).Should((HaveOccurred()))
```
jamiemonserrate, line 36: **Nit** - How come there are extra braces? Those aren't needed right?

jamiemonserrate (approved):
Minor non blocking comments. LGTM

PR #319, opened 2022-01-17
Add a script to run e2e test

### LINKED ISSUE #231: e2e tests exercise the installer
At the moment, we skip running the installer in our e2e tests. This is because the containers need to run in privileged mode, and they end up making kernel changes on the system they are running on. 

We should have the tests spin up a VM, and then run the BYOH host containers in this VM, so that the users system is untouched. Once we do that, we should enable the installer to run in our e2e tests.

anusha94:
- the make target warns the user of the potential kernel config changes if
  run locally
- if the user selects yes, then proceed to run only quickstart e2e
  test
- we will run the test by default in CI

Quoting the warning info from here - 
https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/072890aadd636575b119c37d53d28d96983baf26/agent/installer/internal/algo/installer.go#L30-L36

Fixes #231

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -26,4 +26,4 @@ jobs:
       run: sudo sysctl -w net.netfilter.nf_conntrack_max=131072
 
     - name: Run PR-Blocking e2e tests
-      run: GINKGO_FOCUS="\[PR-Blocking\]" make test-e2e
+      run: yes | make test-e2e
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -26,4 +26,4 @@ jobs:
       run: sudo sysctl -w net.netfilter.nf_conntrack_max=131072
 
     - name: Run PR-Blocking e2e tests
-      run: yes | make test-e2e
+      run: yes | GINKGO_FOCUS="\[PR-Blocking\]" make test-e2e
--- a/Makefile
+++ b/Makefile
@@ -125,7 +125,7 @@ prepare-byoh-docker-host-image:
 prepare-byoh-docker-host-image-dev:
 	docker build test/e2e -f docs/BYOHDockerFileDev -t ${BYOH_BASE_IMG_DEV}
 
-test-e2e: docker-build prepare-byoh-docker-host-image $(GINKGO) cluster-templates-e2e ## Run the end-to-end tests
+test-e2e: take-user-input docker-build prepare-byoh-docker-host-image $(GINKGO) cluster-templates-e2e ## Run the end-to-end tests
 	CONTROL_PLANE_ENDPOINT_IP=172.18.10.151 $(GINKGO) -v -trace -tags=e2e -focus="$(GINKGO_FOCUS)" $(_SKIP_ARGS) -nodes=$(GINKGO_NODES) --noColor=$(GINKGO_NOCOLOR) $(GINKGO_ARGS) test/e2e -- \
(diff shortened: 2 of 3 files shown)


NilanjanDaw (approved):
lgtm!

scripts/test_e2e.sh
```diff
@@ -0,0 +1,33 @@
...
+###############################################################
+
+** WARNING **
+
+Running e2e tests locally will change the following host config
+- enable the kernel modules: overlay & bridge network filter
+- create a systemwide config that will enable those modules at boot time
+- enable ipv4 & ipv6 forwarding
+- create a systemwide config that will enable the forwarding at boot time
+- realod the sysctl with the applied config changes so the changes can take effect without restarting
```
jamiemonserrate, line 19:
```suggestion
- reload the sysctl with the applied config changes so the changes can take effect without restarting
```

scripts/test_e2e.sh
```diff
@@ -0,0 +1,33 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+
+set -o errexit
+set -o pipefail
+
+echo '''
+###############################################################
+
+** WARNING **
```
jamiemonserrate, line 12:
```suggestion
** WARNING ** - These tests modify system settings - and do **NOT** revert them at the end of the test. A list of changes can be found below. We **highly** recommend running these tests in a VM. 
```

scripts/test_e2e.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+# Copyright 2021 VMware, Inc. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+
+set -o errexit
+set -o pipefail
+
+echo '''
```
jamiemonserrate, line 9: Shouldn't this be done in the `test-e2e` make task? Otherwise we run the risk of someone running `make test-e2e`, and them skipping this warning. No?

anusha94: I see your point. Was checking out other repos - calling a make task from the script seems to be the pattern

https://github.com/kubernetes-sigs/cluster-api/blob/18c7ec867f7e8eb690aa7708afe6bbc27fec1498/scripts/ci-e2e.sh#L100

https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/f7c3a7495d26e96fb50d83ff0d1994016e3b211e/hack/e2e.sh#L110

https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/2969152895a268fdaf022156e3b14777d9bcfd8f/scripts/ci-e2e.sh#L84

How about documenting how to run our e2e tests? No? People won't read you say?
Or call the script from the make task and execute ginkgo from the script? We could do it - and I don't have strong opinions - just doesn't look like what everybody else is doing. But neither do they have tests modifying the state. So. Your call.
@jamiemonserrate

dharmjit: I am more aligned towards having a single way to run e2e tests. Not sure though, but it might be possible to do the same thing in `Makefile`. I just tried below to read the input
```
test-e2e: 
	@read -p "Do you want to proceed [Y/n]?" REPLY; \
	echo "Your reply is ", $$REPLY
```

anusha94: Okay then, warnings in the make target it is!
I'll update the PR.
Thank you!

dharmjit (approved):
Looks good.

jamiemonserrate (approved):
LGTM!

PR #328, opened 2022-01-18
fix bug: Got error: Node \"node02\" is invalid: spec.providerID: Forbidden: node updates may not change providerID except from \"\" to valid 

### LINKED ISSUE #290: Got  error: Node \"node02\" is invalid: spec.providerID: Forbidden: node updates may not change providerID except from \"\" to valid
I met a bug when I do test “Create a workload cluster with 3 cp + 3md”. This test can passed with a clear environment. But in my environment, it reused a lot of times, and trigger this following issue:

The second control plane entering “Provisioning” status forever.

image

I met the following error in byoh-controller-manager.log:

> E1216 08:50:09.084241       1 byomachine_controller.go:272] controller/byomachine "msg"="failed to set node providerID" "error"="Node \"node02\" is invalid: spec.providerID: Forbidden: node updates may not change providerID except from \"\" to valid" "cluster"="byoh-wc" "name"="byoh-wc-control-plane-29fxm" "namespace"="default" "reconciler group"="infrastructure.cluster.x-k8s.io" "reconciler kind"="ByoMachine"
> E1216 08:50:09.085312       1 controller.go:317] controller/byomachine "msg"="Reconciler error" "error"="Node \"node02\" is invalid: spec.providerID: Forbidden: node updates may not change providerID except from \"\" to valid" "name"="byoh-wc-control-plane-29fxm" "namespace"="default" "reconciler group"="infrastructure.cluster.x-k8s.io" "reconciler kind"="ByoMachine"   

After read the code and analyse the log, I figure out the reason of this issue.  This reason is the first time to set “Updating Node with ProviderID” failed with timeout, but actually the node object “node02" already update with spec.ProviderID “byoh://node02/hg7i3d”. 

The content of byoh-controller-manager.log:

> E1216 08:50:03.164717       1 byomachine_controller.go:272] controller/byomachine "msg"="failed to set node providerID" "error"="etcdserver: request timed out" "cluster"="byoh-wc" "name"="byoh-wc-control-plane-29fxm" "namespace"="default" "reconciler group"="infrastructure.cluster.x-k8s.io" "reconciler kind"="ByoMachine"  
> E1216 08:50:03.165896       1 controller.go:317] controller/byomachine "msg"="Reconciler error" "error"="etcdserver: request timed out"  

The yaml output of node "node02":

image

Then every time after that will be report this error in byoh-controller-manager**.log:

> E1216 08:50:09.084241       1 byomachine_controller.go:272] controller/byomachine "msg"="failed to set node providerID" "error"="Node \"node02\" is invalid: spec.providerID: Forbidden: node updates may not change providerID except from \"\" to valid" "cluster"="byoh-wc" "name"="byoh-wc-control-plane-29fxm" "namespace"="default" "reconciler group"="infrastructure.cluster.x-k8s.io" "reconciler kind"="ByoMachine"
> E1216 08:50:09.085312       1 controller.go:317] controller/byomachine "msg"="Reconciler error" "error"="Node \"node02\" is invalid: spec.providerID: Forbidden: node updates may not change providerID except from \"\" to valid" "name"="byoh-wc-control-plane-29fxm" "namespace"="default" "reconciler group"="infrastructure.cluster.x-k8s.io" "reconciler kind"="ByoMachine"

And this error output is because it is not allowed to update the spec.providerID of node object. You can try to delete this field of a node object, then you’ll see this error output. With this error, the control plane entering “Provisioning” status forever.

huchen2021:
Fixes #290

This issue is not easy to reproduce, and I only met once. It is triggered when the following scenario happened:
In the reconcile function of the byomachine:
    1) The first time to set “Updating Node with ProviderID” return failed with timeout, but actually the node object “node02" already update with spec.ProviderID “byoh://node02/hg7i3d”.
    2) Then every time after “Updating Node with ProviderID” return failed with Node “node02” is invalid: spec.providerID: Forbidden: node updates may not change providerID except from “” to valid”
This is because it is not allowed to update the spec.providerID of node object. This error prevent the continuation of reconcile. And If this happened, it need to remove the node from the cluster manually and restart kubelet on it.

I open this issue to track it, and I was thinking of a solution which may fix the issue without manual operation:
Inside the setNodeProviderID function, if field is not empty and equal to providerID, then skip patch.
If not empty and not equal to providerID, then throw a custom error message cannot set providerID on the node.

### DIFF
--- a/controllers/infrastructure/byomachine_controller.go
+++ b/controllers/infrastructure/byomachine_controller.go
@@ -360,7 +360,8 @@ func (r *ByoMachineReconciler) setNodeProviderID(ctx context.Context, remoteClie
 	}
 
 	if node.Spec.ProviderID != "" {
-		match, err := regexp.MatchString("byoh://"+host.Name+"/.+", node.Spec.ProviderID)
+		var match bool
+		match, err = regexp.MatchString("byoh://"+host.Name+"/.+", node.Spec.ProviderID)
 		if err != nil {
 			return "", err
 		}
--- a/controllers/infrastructure/byomachine_controller.go
+++ b/controllers/infrastructure/byomachine_controller.go
@@ -375,7 +375,7 @@ func (r *ByoMachineReconciler) setNodeProviderID(ctx context.Context, remoteClie
 		return "", err
 	}
 
-	node.Spec.ProviderID = fmt.Sprintf("%s%s/%s", providerIDPrefix, host.Name, util.RandomString(providerIDSuffixLength))
+	node.Spec.ProviderID = fmt.Sprintf("%s%s/%s", ProviderIDPrefix, host.Name, util.RandomString(providerIDSuffixLength))
 
 	return node.Spec.ProviderID, helper.Patch(ctx, node)
 }
--- a/controllers/infrastructure/byomachine_controller.go
+++ b/controllers/infrastructure/byomachine_controller.go
@@ -260,19 +260,20 @@ func (r *ByoMachineReconciler) reconcileNormal(ctx context.Context, machineScope
 
 	if machineScope.ByoMachine.Spec.ProviderID == "" {
 		logger.Info("Updating Node with ProviderID")
-		providerID := fmt.Sprintf("%s%s/%s", providerIDPrefix, machineScope.ByoHost.Name, util.RandomString(providerIDSuffixLength))
+
 		remoteClient, err := r.getRemoteClient(ctx, machineScope.ByoMachine)
 		if err != nil {
 			logger.Error(err, "failed to get remote client")
 			return ctrl.Result{}, err
 		}
 
-		err = r.setNodeProviderID(ctx, remoteClient, machineScope.ByoHost, providerID)
+		providerID, err := r.setNodeProviderID(ctx, remoteClient, machineScope.ByoHost, machineScope.ByoHost.Name)
 		if err != nil {
 			logger.Error(err, "failed to set node providerID")
 			r.Recorder.Eventf(machineScope.ByoMachine, corev1.EventTypeWarning, "SetNodeProviderFailed", "Node %s does not exist", machineScope.ByoHost.Name)
 			return ctrl.Result{}, err
 		}
+
 		machineScope.ByoMachine.Spec.ProviderID = providerID
 		machineScope.ByoMachine.Status.Ready = true
 		conditions.MarkTrue(machineScope.ByoMachine, infrav1.BYOHostReady)
@@ -349,22 +350,33 @@ func (r *ByoMachineReconciler) ClusterToByoMachines(logger logr.Logger) handler.
 
 // setNodeProviderID patches the provider id to the node using
 // client pointing to workload cluster
-func (r *ByoMachineReconciler) setNodeProviderID(ctx context.Context, remoteClient client.Client, host *infrav1.ByoHost, providerID string) error {
+func (r *ByoMachineReconciler) setNodeProviderID(ctx context.Context, remoteClient client.Client, host *infrav1.ByoHost, byoHostName string) (string, error) {
 	node := &corev1.Node{}
 	key := client.ObjectKey{Name: host.Name, Namespace: host.Namespace}
 	err := remoteClient.Get(ctx, key, node)
-
 	if err != nil {
-		return err
+		return "", err
+	}
+
+	providerIDLen, providerIDPrefixLen := len(node.Spec.ProviderID), len(providerIDPrefix)
+	if providerIDLen > 0 {
+		if providerIDLen > providerIDPrefixLen+providerIDSuffixLength {
+			providerByohostName := node.Spec.ProviderID[providerIDPrefixLen : providerIDLen-providerIDSuffixLength-1]
+			if providerByohostName == byoHostName {
+				return node.Spec.ProviderID, nil
+			}
+			return "", fmt.Errorf("can't set provider to %s, since it already is %s", byoHostName, providerByohostName)
+		}
+		return "", errors.New("invalid format for node.Spec.ProviderID")
 	}
 	helper, err := patch.NewHelper(node, remoteClient)
 	if err != nil {
-		return err
+		return "", err
 	}
 
-	node.Spec.ProviderID = providerID
+	node.Spec.ProviderID = fmt.Sprintf("%s%s/%s", providerIDPrefix, byoHostName, util.RandomString(providerIDSuffixLength))
 
-	return helper.Patch(ctx, node)
+	return node.Spec.ProviderID, helper.Patch(ctx, node)
 }
 
 func (r *ByoMachineReconciler) getRemoteClient(ctx context.Context, byoMachine *infrav1.ByoMachine) (client.Client, error) {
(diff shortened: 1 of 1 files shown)


controllers/infrastructure/byomachine_controller.go
```diff
@@ -349,22 +350,33 @@ func (r *ByoMachineReconciler) ClusterToByoMachines(logger logr.Logger) handler.
...
 	node := &corev1.Node{}
 	key := client.ObjectKey{Name: host.Name, Namespace: host.Namespace}
 	err := remoteClient.Get(ctx, key, node)
-
 	if err != nil {
-		return err
+		return "", err
+	}
+
+	providerIDLen, providerIDPrefixLen := len(node.Spec.ProviderID), len(providerIDPrefix)
```
anusha94, line 361: Can we simplify this to 
```
if node.Spec.ProviderID != "" {
    if node.Spec.ProviderID == providerID {
        // do nothing and return
    }
    // return with err
}
// continue to set node.Spec.ProviderID == providerID
```

anusha94 (commented):
Also good to add tests in the same PR.

controllers/infrastructure/byomachine_controller.go
huchen2021: No, actually, providerID contains random strings, and we don't know what's that. What we need is byohost name.

anusha94: Okay I see now. Since `providerID` contains random string, we cannot do a full match.
In this case, can we define a helper function `getByoHostNamefromProviderID` that will accept a `providerID` and return the `byohost name`.
This helper function could probably reside in a file called `controllerutils.go` inside https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/tree/main/controllers/infrastructure?

Also, since we know the `providerID` pattern `byoh://host-name/string`, we should be able to define regex to extract this value, this will be more elegant than slicing string based on length.

controllers/infrastructure/byomachine_controller.go
```diff
@@ -260,19 +260,20 @@ func (r *ByoMachineReconciler) reconcileNormal(ctx context.Context, machineScope
...
-		providerID := fmt.Sprintf("%s%s/%s", providerIDPrefix, machineScope.ByoHost.Name, util.RandomString(providerIDSuffixLength))
+
 		remoteClient, err := r.getRemoteClient(ctx, machineScope.ByoMachine)
 		if err != nil {
 			logger.Error(err, "failed to get remote client")
 			return ctrl.Result{}, err
 		}
 
-		err = r.setNodeProviderID(ctx, remoteClient, machineScope.ByoHost, providerID)
+		providerID, err := r.setNodeProviderID(ctx, remoteClient, machineScope.ByoHost, machineScope.ByoHost.Name)
```
anusha94, line 270: You are already passing `machineScope.ByoHost`, no need to pass `machineScope.ByoHost.Name`

huchen2021: Will fix it.

controllers/infrastructure/byomachine_controller.go
huchen2021: Thanks to your comment, and I have a better idea. No need to  define a helper function getByoHostNamefromProviderID. It will be more simpler.

huchen2021:
> Also good to add tests in the same PR.

I am afraid it is not easy to do this. Because this bug is very hard to reproduce. And it doesn't have specified scenario to trigger it neither.

controllers/infrastructure/byomachine_controller.go
```diff
@@ -260,19 +261,20 @@ func (r *ByoMachineReconciler) reconcileNormal(ctx context.Context, machineScope
...
-		providerID := fmt.Sprintf("%s%s/%s", providerIDPrefix, machineScope.ByoHost.Name, util.RandomString(providerIDSuffixLength))
+
 		remoteClient, err := r.getRemoteClient(ctx, machineScope.ByoMachine)
 		if err != nil {
 			logger.Error(err, "failed to get remote client")
 			return ctrl.Result{}, err
 		}
 
-		err = r.setNodeProviderID(ctx, remoteClient, machineScope.ByoHost, providerID)
+		providerID, err := r.setNodeProviderID(ctx, remoteClient, machineScope.ByoHost)
```
anusha94, line 271: Trying to keep the change minimal. Any reason why `providerID` computation is moved inside `setNodeProviderID`?

huchen2021: That's because, providerID may don't need computation, it may return by setNodeProviderID function under some situation.

anusha94:
> I am afraid it is not easy to do this. Because this bug is very hard to reproduce. And it doesn't have specified scenario to trigger it neither.

Agreed. But we know the scenario under which this error will occur. We can write our setup code accordingly. Can you use this node object and set a `providerID` that will match the regex condition? Then we can unit test the new code.
https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/2b30d6fd0d11052422d6d78e0d57c9a83db95969/controllers/infrastructure/byomachine_controller_test.go#L183

controllers/infrastructure/byomachine_controller.go
anusha94: Ah okay, right. Thanks!

huchen2021:
> > I am afraid it is not easy to do this. Because this bug is very hard to reproduce. And it doesn't have specified scenario to trigger it neither.
> 
> Agreed. But we know the scenario under which this error will occur. We can write our setup code accordingly. Can you use this node object and set a `providerID` that will match the regex condition? Then we can unit test the new code.
> 
> https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/2b30d6fd0d11052422d6d78e0d57c9a83db95969/controllers/infrastructure/byomachine_controller_test.go#L183

 [issue-335](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/issues/335) is created to track this requirement. let's focus code fix on this RP. What do you think? @anusha94

controllers/infrastructure/byomachine_controller.go
```diff
@@ -349,22 +350,34 @@ func (r *ByoMachineReconciler) ClusterToByoMachines(logger logr.Logger) handler.
...
 	err := remoteClient.Get(ctx, key, node)
-
 	if err != nil {
-		return err
+		return "", err
+	}
+
+	if node.Spec.ProviderID != "" {
+		var match bool
+		match, err = regexp.MatchString("byoh://"+host.Name+"/.+", node.Spec.ProviderID)
```
dharmjit, line 363: Suggestion: No strong opinions but I find `fmt.Sprintf` more readable instead of concatenation. wdyt?

anusha94:
> issue-335 is created to track this requirement. let's focus code fix on this RP. What do you think?

Sure @huchen2021 
But having the test in the same PR makes the code changes more easily understandable. Maybe we can do that in future, should be fine for this PR.

huchen2021:
> > issue-335 is created to track this requirement. let's focus code fix on this RP. What do you think?
> 
> Sure @huchen2021 But having the test in the same PR makes the code changes more easily understandable. Maybe we can do that in future, should be fine for this PR.

Thanks for understanding.

controllers/infrastructure/byomachine_controller.go
huchen2021: ok.

dharmjit (approved):
lgtm

PR #333, opened 2022-01-18
Bugfix kubeadm cleanup

### LINKED ISSUE #298: Host cleanup fails with kubeadm not found error
- follow the getting_started guide to create workload cluster on VMs
- the k8s components installation failed due to some reason
- now the reset host method kicks in attempting to execute `kubeadm reset --force`
- since the k8s components were not installed, `kubeadm` is not installed, therefore `kubeadm not found` error

- host should be reset and back to `Waiting for MachineRef state`

In the `hostCleanup` function, `resetNode` should be called only if `K8sComponentsInstallationSucceeded Condition is true` --> this means kubeadm will be present on the machine.
https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/970c5de3c5100f8fb93b6ac85516e6ef735261ec/agent/reconciler/host_reconciler.go#L189-L195

NilanjanDaw:
Fix node clean-up logic to skip node reset in case the k8s installation fails

Fixes #298

### DIFF
--- a/agent/reconciler/host_reconciler.go
+++ b/agent/reconciler/host_reconciler.go
@@ -209,7 +209,17 @@ func (r *HostReconciler) hostCleanUp(ctx context.Context, byoHost *infrastructur
 	}
 	conditions.MarkFalse(byoHost, infrastructurev1beta1.K8sNodeBootstrapSucceeded, infrastructurev1beta1.K8sNodeAbsentReason, clusterv1.ConditionSeverityInfo, "")
 
-	err := r.removeAnnotations(ctx, byoHost)
+	err := r.removeSentinelFile(ctx, byoHost)
+	if err != nil {
+		return err
+	}
+
+	err = r.deleteEndpointIP(ctx, byoHost)
+	if err != nil {
+		return err
+	}
+
+	err = r.removeAnnotations(ctx, byoHost)
 	if err != nil {
 		return err
 	}
@@ -263,7 +273,7 @@ func (r *HostReconciler) installK8sComponents(ctx context.Context, byoHost *infr
 
 func (r *HostReconciler) uninstallk8sComponents(ctx context.Context, byoHost *infrastructurev1beta1.ByoHost) error {
 	logger := ctrl.LoggerFrom(ctx)
-	
+
 	bundleRegistry := byoHost.GetAnnotations()[infrastructurev1beta1.BundleLookupBaseRegistryAnnotation]
 	k8sVersion := byoHost.GetAnnotations()[infrastructurev1beta1.K8sVersionAnnotation]
 	byohBundleTag := byoHost.GetAnnotations()[infrastructurev1beta1.BundleLookupTagAnnotation]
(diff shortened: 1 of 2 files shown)


agent/reconciler/host_reconciler.go
```diff
@@ -206,6 +212,7 @@ func (r *HostReconciler) hostCleanUp(ctx context.Context, byoHost *infrastructur
 		}
 		err = bundleInstaller.Uninstall(k8sVersion, byohBundleTag)
 		if err != nil {
+			logger.Info(err.Error())
```
anusha94, line 215: We are already returning `err` (which will be logged at some point in the error handling chain), so no need to log separately.

agent/reconciler/host_reconciler.go
```diff
@@ -189,9 +189,15 @@ func (r *HostReconciler) cleank8sdirectories(ctx context.Context) error {
 func (r *HostReconciler) hostCleanUp(ctx context.Context, byoHost *infrastructurev1beta1.ByoHost) error {
 	logger := ctrl.LoggerFrom(ctx)
 	logger.Info("cleaning up host")
-	err := r.resetNode(ctx, byoHost)
-	if err != nil {
-		return err
+
+	k8sComponentsInstallationSucceeded := conditions.Get(byoHost, infrastructurev1beta1.K8sComponentsInstallationSucceeded)
+	if k8sComponentsInstallationSucceeded != nil && k8sComponentsInstallationSucceeded.Status == corev1.ConditionTrue {
```
anusha94, line 194: I think this is good enough code to break it down into multiple functions. 
`hostCleanup` is doing 3 things:
1. execute kubeadm reset
2. uninstall k8s components
3. remove a bunch of annotations

Both 1 & 2 can be skipped if `k8sComponentsInstallationSucceeded` is not true

@dharmjit Do you see any concern with this? I don't think there is anything as partial `Install` successful. So I'm okay to skip uninstall if this condition is not true.

dharmjit: @anusha94 I agree as `bootstrapK8sNode` comes after `installK8sComponents`, It is okay to skip both 1 and 2 if `k8sComponentsInstallationSucceeded` is not true.

agent/reconciler/host_reconciler.go
```diff
@@ -302,3 +260,67 @@ func (r *HostReconciler) installK8sComponents(ctx context.Context, byoHost *infr
...
+	}
+	if IP, ok := byoHost.Annotations[infrastructurev1beta1.EndPointIPAnnotation]; ok {
+		network, err := vip.NewConfig(IP, registration.LocalHostRegistrar.ByoHostInfo.DefaultNetworkInterfaceName, false)
+		if err == nil {
+			err := network.DeleteIP()
+			if err != nil {
+				return err
+			}
+		}
+	}
```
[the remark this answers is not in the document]
NilanjanDaw: Will it better if we rename this to `removeMetadata` or `removeClusterMetadata`?

agent/reconciler/host_reconciler.go
NilanjanDaw: Have realigned `hostCleanup` to now skip both node reset and k8s component uninstallation if `k8sComponentsInstallationSucceeded` is not true

agent/reconciler/host_reconciler.go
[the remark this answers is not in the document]
dharmjit: `removeMetadata` sounds okay to me. @anusha94 thoughts??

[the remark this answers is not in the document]
anusha94: > This code doesn't align with the method name removeAnnotations

+1

I would not mind having 2 more functions - `removeSentinelFile()` and `deleteEndpointIP()` that will be called from `hostCleanUp` function.
Or you can choose to keep this inside `hostCleanUp` itself like @dharmjit suggested.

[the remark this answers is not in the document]
anusha94: `removeAnnotations` is a better name IMO - for annotations related stuff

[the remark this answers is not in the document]
NilanjanDaw: I think adding two more functions `removeSentinelFile()` and `deleteEndpointIP()` along with `removeAnnotations()` sounds good. It would keep the `hostCleanup` function clean

agent/reconciler/host_reconciler.go
```diff
@@ -189,72 +189,40 @@ func (r *HostReconciler) cleank8sdirectories(ctx context.Context) error {
...
 
-	// Remove the bundle tag annotation
-	delete(byoHost.Annotations, infrastructurev1beta1.BundleLookupTagAnnotation)
+	err = r.deleteEndpointIP(ctx, byoHost)
+	if err != nil {
+		return err
+	}
 
+	err = r.removeAnnotations(ctx, byoHost)
```
anusha94, line 222: There is no way `removeAnnotations` is going to return an error. Right? Wondering if we can get rid of the below `if` check for `err`

NilanjanDaw: At the moment no, I thought about removing it but kept it in case we add something in future to `removeAnnotations` which returns an error.

anusha94: I think its good to write error handling when we have error being returned. We can remove it now. So that code becomes a little bit more tidy.

NilanjanDaw: Cool sounds good

anusha94 (approved):
lgtm

dharmjit (approved):
lgtm

PR #346, opened 2022-01-24
Add version flag to host agent

### LINKED ISSUE #278: Add --version flag to host agent binary
It will be useful to provide a `--version` flag in the host-agent binary as we will have multiple agent releases so that the user can know the version being used. The output could consist of information in the below format

```
$ byoh-hostagent-linux-amd64 --version
Version: v0.1.0_alpha.2
BuildDate: 
...
```

raunaksingwi:
Add version flag to the host agent so that the user can know the version being used.
The output shows the version of the binary and the time it was built.

Fixes #278 

**Additional information**
Used ldflags to embed the version number and time at build time in the binary.

The version number has to be passed as a variable to the Makefile. It will be passed by the GH action while the build is triggered.
When the version is not passed, it takes the default as "dev"


anusha94 (approved):
lgtm

agent/version/version_test.go
```diff
@@ -0,0 +1,107 @@
...
+				BuildDate: version.BuildDate,
+				GoVersion: runtime.Version(),
+				Compiler:  runtime.Compiler,
+				Platform:  fmt.Sprintf("%s/%s", runtime.GOOS, runtime.GOARCH),
+			}
+			Expect(version.Get()).Should(Equal(expected))
+		})
+	})
+
+	Context("When version is set", func() {
```
dharmjit, line 60: Suggestion: Should we have one more context under this for `when the git tag is not proper` for e.g. 1.0, etc.

dharmjit (commented):
IMO, we need a consistent organization for `HostAgent` flags related code. Currently some of the code(help, labels flag) is inside `main` package and this PR introduces a new package `version`. How could we make it more cleaner? thoughts @anusha94 @raunaksingwi @huchen2021

anusha94:
> IMO, we need a consistent organization for `HostAgent` flags related code. Currently some of the code(help, labels flag) is inside `main` package and this PR introduces a new package `version`. How could we make it more cleaner? thoughts @anusha94 @raunaksingwi @huchen2021

Agreed. All flags are all over the place now. @raunaksingwi Mind creating a new issue to address this so that this PR is unblocked?

anusha94 (approved):

dharmjit (approved):
lgtm

Issue #340, opened 2022-01-19
Improve codecov for agent/reconciler


anusha94:
At the moment, we have 63% coverage for reconciler. Identify more test cases to improve the coverage. To look at the detailed codecov report - https://app.codecov.io/gh/vmware-tanzu/cluster-api-provider-bringyourownhost

Tasks -
- [x] uncomment unit test for uninstall flow https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/1fab38acf1d3a4fbe27ca147f694ee7f028bf712/agent/reconciler/reconciler_test.go#L431-L449
- [x] uncomment unit test for install flow https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/1fab38acf1d3a4fbe27ca147f694ee7f028bf712/agent/reconciler/reconciler_test.go#L178-L198
This will involve some changes to initializing installer in main.go and mocking installer's Install & Uninstall methods

- [x] #393

anusha94:
This issue can be worked upon after completing #221 and #254

NilanjanDaw:
Need an integration test reconciler to cover `bundle_downlaoder.go` especially `Download()` and `downloadByImgpkg()`

NilanjanDaw:
All tasks are done.

PR #356, opened 2022-01-27
Add pre-requisite package precheck to byoh-agent

### LINKED ISSUE #254: Pre-requisite check on Host-Agent start
When using an `Ubuntu 20.04 VM` (with no additional installation of packages), applying the cluster-template proceeds until the k8s installation steps. It later complains of the missing installation required packages like `socat, ebtables, conntrack, ethtool`.

The requirements are already documented [here](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/main/docs/local_dev.md#pre-requisites-1)
But if the user has not followed the steps to install these, host-agent on start should check for the pre-requisites (all required packages, os support etc.) and exit on failure.
This will save effort in registering a byohost that does not meet basic requirements and will go into a trying to `install / uninstall` reconcile loop. Manual intervention should not be needed.

raunaksingwi:
Currently we only support byoh-agent on Ubuntu 20.04.x
Also when the user doesn't choose to skip installation for kubernetes
component binaries, following packages are expected to be pre-installed:
socat, ebtables, ethtool, conntrack
Add precheck to verify the above two conditions

Fixes #254

### DIFF
--- a/agent/installer/installer.go
+++ b/agent/installer/installer.go
@@ -6,8 +6,10 @@ package installer
 import (
 	"errors"
 	"fmt"
-	"os/exec"
+	"runtime"
 	"strings"
+
+	utilsexec "k8s.io/utils/exec"
 
 	"github.com/go-logr/logr"
 	"github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/agent/installer/internal/algo"
@@ -90,15 +92,19 @@ func (bd *bundleDownloader) DownloadOrPreview(os, k8s, tag string) error {
 }
 
 func ckeckPreRequsitePackages() error {
-	unavailablePackages := []string{}
-	for _, pkgName := range preRequisitePackages {
-		_, err := exec.Command("dpkg-query", "-W", pkgName).Output()
-		if err != nil {
-			unavailablePackages = append(unavailablePackages, pkgName)
+	if runtime.GOOS == "linux" {
+		unavailablePackages := []string{}
+		execr := utilsexec.New()
+		for _, pkgName := range preRequisitePackages {
+			_, err := execr.LookPath(pkgName)
+			if err != nil {
+				unavailablePackages = append(unavailablePackages, pkgName)
+			}
 		}
-	}
-	if len(unavailablePackages) != 0 {
-		return fmt.Errorf("required package(s): %s not found", unavailablePackages)
+		if len(unavailablePackages) != 0 {
+			return fmt.Errorf("required package(s): %s not found", unavailablePackages)
+		}
+		return nil
 	}
 	return nil
 }
--- a/agent/installer/installer.go
+++ b/agent/installer/installer.go
@@ -130,7 +130,7 @@ func runPrechecks(logger logr.Logger, os string) bool {
 	}
 
 	// verify that packages are available when user has chosen to install kubernetes binaries
-	err = preckeckPreRequsitPackages()
+	err = ckeckPreRequsitePackages()
 	if err != nil {
 		logger.Error(err, "Failed pre-requisite packages precheck")
 		precheckSuccessful = false
--- a/agent/installer/installer.go
+++ b/agent/installer/installer.go
@@ -98,8 +98,7 @@ func ckeckPreRequsitePackages() error {
 		}
 	}
 	if len(unavailablePackages) != 0 {
-		errMsg := fmt.Sprintf("Required package(s): %s not found", unavailablePackages)
-		return errors.New(errMsg)
+		return fmt.Errorf("required package(s): %s not found", unavailablePackages)
 	}
 	return nil
 }
--- a/agent/installer/installer.go
+++ b/agent/installer/installer.go
@@ -132,7 +132,7 @@ func runPrechecks(logger logr.Logger, os string) bool {
 	// verify that packages are available when user has chosen to install kubernetes binaries
 	err = preckeckPreRequsitPackages()
 	if err != nil {
-		logger.Error(err, "Failed pre-requisit packages precheck")
+		logger.Error(err, "Failed pre-requisite packages precheck")
 		precheckSuccessful = false
 	}
 	return precheckSuccessful
--- a/agent/installer/installer.go
+++ b/agent/installer/installer.go
@@ -6,10 +6,7 @@ package installer
 import (
 	"errors"
 	"fmt"
-	"runtime"
 	"strings"
-
-	utilsexec "k8s.io/utils/exec"
 
 	"github.com/go-logr/logr"
 	"github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/agent/installer/internal/algo"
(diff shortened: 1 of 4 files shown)


anusha94:
Apologies for reviewing while still in Draft state. But I took a quick look -
I agree with the checks but I prefer the checks to be present in the installer package rather than in main.
Why?
1. these requirements are for the installer and not agent
2. any future changes to the installer in terms of OS / packages etc can all be done in one place

agent/installer/installer_test.go
```diff
@@ -128,6 +128,52 @@ var _ = Describe("Byohost Installer Tests", func() {
 			Expect(func() { NewPreviewInstaller("Ubuntu_20.04.3_x86-64", nil) }).NotTo(Panic())
 		})
 	})
+	Context("When supported OS version is passed for precheck", func() {
+		It("Should not return error", func() {
+			err := supportedOSversion("Ubuntu_20.04.1_x86-64")
```
anusha94, line 133: Currently this test is marked with `nolint` which needs to be removed in future. Can we avoid testing private functions? Can it not be via the public method that in turn calls this function?
Because the agent code has no direct access to this function.

agent/installer/installer.go
```diff
@@ -85,6 +91,53 @@ func (bd *bundleDownloader) DownloadOrPreview(os, k8s, tag string) error {
...
+		return err
+	}
+	if !matched {
+		errMsg := fmt.Sprintf("Unsupported OS. Note: Currently supported OS is %s", SupportedOS)
+		return errors.New(errMsg)
+	}
+	return nil
+}
+
+func preckeckPreRequsitPackages() error {
```
anusha94, line 106:
```suggestion
func ckeckPreRequsitePackages() error {
```

agent/installer/installer.go
```diff
@@ -85,6 +91,53 @@ func (bd *bundleDownloader) DownloadOrPreview(os, k8s, tag string) error {
...
+	err := supportedOSversion(os)
+	if err != nil {
+		logger.Error(err, "Failed supported OS version precheck")
+		precheckSuccessful = false
+	}
+
+	// verify that packages are available when user has chosen to install kubernetes binaries
+	err = preckeckPreRequsitPackages()
+	if err != nil {
+		logger.Error(err, "Failed pre-requisit packages precheck")
```
anusha94, line 135:
```suggestion
		logger.Error(err, "Failed pre-requisite packages precheck")
```

agent/installer/installer.go
```diff
@@ -85,6 +91,53 @@ func (bd *bundleDownloader) DownloadOrPreview(os, k8s, tag string) error {
 	return bd.Download(os, k8s, tag)
 }
 
+func supportedOSversion(os string) error {
```
anusha94, line 94: Have you looked at https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/f285e157aa01b501aa7c54a48374c1d25d7c772d/agent/installer/installer.go#L116-L118

Looks like this is also kind of validating what OS is supported. Maybe we can leverage this? Or have only one place where we check this (i.e combine both code)

anusha94: cc @jeuxdemains

raunaksingwi: Looking at it now

raunaksingwi: Ok, with this we don't need another OS precheck. Hence removed the check that I had added.

agent/installer/installer.go
```diff
@@ -85,6 +91,53 @@ func (bd *bundleDownloader) DownloadOrPreview(os, k8s, tag string) error {
...
+		errMsg := fmt.Sprintf("Unsupported OS. Note: Currently supported OS is %s", SupportedOS)
+		return errors.New(errMsg)
+	}
+	return nil
+}
+
+func ckeckPreRequsitePackages() error {
+	unavailablePackages := []string{}
+	for _, pkgName := range PreRequisitePackages {
+		_, err := exec.Command("dpkg-query", "-W", pkgName).Output()
```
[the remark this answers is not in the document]
anusha94: cc @georgievaVMW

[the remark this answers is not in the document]
jeuxdemains: The code is OS/Platform agnostic, except for this line I guess. We need a better way to do this check. We can either use the current installer approach and provide a matrix of <package type> to <package manager cmd> mapping or provide the package type information externally. Both have pros and cons. Any proposals are very welcome.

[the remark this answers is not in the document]
jamiemonserrate: I am slightly hesitant, since this feels like a duplicate of the kubeadm preflight checks. But I understand that we don't have kubeadm when the host-agent starts, and if we wait until we have kubeadm - its too late.

Can we either reuse what kubeadm does? Or maybe follow inspiration from them? https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/preflight/checks.go#L1032-L1043

agent/installer/installer.go
```diff
@@ -24,6 +26,8 @@ const (
 	ErrBundleUninstall   = Error("Error uninstalling bundle")
 )
 
+var PreRequisitePackages = []string{"socat", "ebtables", "ethtool", "conntrack"}
```
jamiemonserrate, line 29: oohhh.. I see this is public so that we can override to test it? I'm personally not a fan. I would rather keep this private - since we don't really want consumers to be able to override this. Right? 

If we do, then I would rather inject this via a constructor. 

For testing, this might be hard (and maybe even pointless) to unit test. Might be easier to write an integration / e2e test of sorts (something that runs in a container).

raunaksingwi: The e2e test already checks this.
So I will make it private and remove the unit tests.

agent/installer/installer.go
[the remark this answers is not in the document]
raunaksingwi: Wow, this was very helpful.
Fixed.

.github/workflows/e2e.yml
```diff
@@ -26,4 +26,4 @@ jobs:
       run: sudo sysctl -w net.netfilter.nf_conntrack_max=131072
 
     - name: Run PR-Blocking e2e tests
-      run: GINKGO_FOCUS="\[PR-Blocking\]" make test-e2e
+      run: yes | GINKGO_FOCUS="\[PR-Blocking\]" make test-e2e
```
anusha94, line 29: Why is this listed as a change? We already have this on `main`. Maybe something wrong with rebase?
https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/2fc856d5f122fad4c8ddc95d023a6f0e14c16059/.github/workflows/e2e.yml#L28-L29

Makefile
```diff
@@ -137,6 +137,32 @@ cluster-templates: kustomize cluster-templates-v1beta1
 cluster-templates-e2e: kustomize
 	$(KUSTOMIZE) build $(BYOH_TEMPLATES)/v1beta1/templates/e2e --load_restrictor none > $(BYOH_TEMPLATES)/v1beta1/templates/e2e/cluster-template.yaml
 
+define WARNING
```
anusha94, line 140: same. Needs rebase from main.

agent/installer/installer.go
```diff
@@ -85,6 +91,36 @@ func (bd *bundleDownloader) DownloadOrPreview(os, k8s, tag string) error {
 	return bd.Download(os, k8s, tag)
 }
 
+func ckeckPreRequsitePackages() error {
```
anusha94, line 94: Being a bit pedantic, can we move these 2 functions to something like `checks.go`? Don't want to bloat the `installer.go` file

anusha94:
The 2 new functions introduced have no tests at the moment. They will be covered when we write integration tests as part of this issue - https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/issues/340

agent/installer/installer.go
```diff
@@ -24,6 +25,8 @@ const (
 	ErrBundleUninstall   = Error("Error uninstalling bundle")
 )
 
+var preRequisitePackages = []string{"socat", "ebtables", "ethtool", "conntrack"}
```
pshail, line 28: futurescope: @anusha94 / @raunaksingwi if we plug this list out i feel there is potential of a generic lib that can be used across projects 😄

pshail (approved):
LGTM

pshail:
@anusha94 are we planning to use #340 as long running issue to track all these misses (new/old)?

anusha94:
> @anusha94 are we planning to use #340 as long running issue to track all these misses (new/old)?

Yes. Long running in the sense we intend to work on that in the next couple weeks. We had no integration tests for the agent and were waiting on this PR https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/pull/316
Now that its merged, we will proceed with writing tests.

This PR is doing system level checks and mocking them for writing unit tests seems like a stretch. For our agent integration tests, we will use containers as hosts, so hoping this newly added code path will be tested then.

agent/installer/installer.go
anusha94: @pshail Can you say more? Plug out where? Coming from a config file of sorts?

pshail: Yup - basically plug-out the requisite packages

pshail:
futurescope: @anusha94 we should ideally have some way of version locking these pre-requisite deps, the more it think its a strong case for a extracted lib

dharmjit (approved):

PR #358, opened 2022-01-31
Code improvements

pshail:
- Fix go code quality issue
  - Code style issues
  - Declaration redundancies
  - Probable bugs
- Add documentations for exported functions
- Add TODO for unused code for cleanup
- General grammar/typo issues

Signed-off-by: Shailesh Pant <7968042+pshail@users.noreply.github.com>

### DIFF
--- a/agent/cloudinit/file_writer.go
+++ b/agent/cloudinit/file_writer.go
@@ -62,7 +62,6 @@ func (w FileWriter) WriteToFile(file *Files) error {
 		return err
 	}
 
-	defer f.Close()
 	_, err = f.WriteString(file.Content)
 	if err != nil {
 		return err
@@ -114,5 +113,5 @@ func (w FileWriter) WriteToFile(file *Files) error {
 		}
 	}
 
-	return nil
+	return f.Close()
 }
--- a/agent/host_agent_suite_test.go
+++ b/agent/host_agent_suite_test.go
@@ -18,7 +18,6 @@ import (
 	infrastructurev1beta1 "github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/apis/infrastructure/v1beta1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/client-go/rest"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/envtest"
@@ -27,10 +26,8 @@ import (
 )
 
 var (
-	err                   error
 	pathToHostAgentBinary string
 	kubeconfigFile        *os.File
-	cfg                   *rest.Config
 	k8sClient             client.Client
 	tmpFilePrefix         = "kubeconfigFile-"
 	testEnv               *envtest.Environment
@@ -39,6 +36,14 @@ var (
 func TestHostAgent(t *testing.T) {
 	RegisterFailHandler(Fail)
 	RunSpecs(t, "Agent Suite")
+}
+
+func getKubeConfig() *os.File {
+	return kubeconfigFile
+}
+
+func setKubeConfig(kubeConfig *os.File) {
+	kubeconfigFile = kubeConfig
 }
 
 var _ = BeforeSuite(func() {
@@ -51,7 +56,7 @@ var _ = BeforeSuite(func() {
 		ErrorIfCRDPathMissing: true,
 	}
 
-	cfg, err = testEnv.Start()
+	cfg, err := testEnv.Start()
 	Expect(err).NotTo(HaveOccurred())
 	Expect(cfg).NotTo(BeNil())
 
@@ -67,6 +72,7 @@ var _ = BeforeSuite(func() {
 	Expect(err).NotTo(HaveOccurred())
 
 	k8sClient, err = client.New(cfg, client.Options{Scheme: scheme})
+	Expect(err).NotTo(HaveOccurred())
 	writeKubeConfig()
 
 	pathToHostAgentBinary, err = gexec.Build("github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/agent")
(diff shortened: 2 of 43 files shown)


test/e2e/md_scale_test.go
```diff
@@ -62,16 +62,18 @@ var _ = Describe("When testing MachineDeployment scale out/in", func() {
...
-			output, byohostContainerID, err := setupByoDockerHost(ctx, clusterConName, byoHostName, namespace.Name, dockerClient, bootstrapClusterProxy)
-			allbyohostContainerIDs = append(allbyohostContainerIDs, byohostContainerID)
-			Expect(err).NotTo(HaveOccurred())
-
-			// read the log of host agent container in backend, and write it
-			agentLogFile := fmt.Sprintf("/tmp/host-agent-%d.log", i)
-			f := WriteDockerLog(output, agentLogFile)
-			defer f.Close()
-			allAgentLogFiles = append(allAgentLogFiles, agentLogFile)
+			func() {
```
anusha94, line 65: Is this a common pattern to define anonymous function inside for loop? I don't understand the difference / improvement from what it was earlier.

test/e2e/e2e_test.go
```diff
@@ -128,9 +138,21 @@ var _ = Describe("When BYOH joins existing cluster [PR-Blocking]", func() {
 			}
 		}
 
-		os.Remove(agentLogFile1)
-		os.Remove(agentLogFile2)
-		os.Remove(ReadByohControllerManagerLogShellFile)
-		os.Remove(ReadAllPodsShellFile)
+		err := os.Remove(agentLogFile1)
+		if err != nil {
```
anusha94, line 142: I don't think we need `if` statements. Directly we can assert `Expect(err).NotTo(HaveOccurred())`

anusha94: Same for all `err != nil` checks in test files

anusha94 (commented):
Thanks for this cleanup!

anusha94:
oops. Just saw it is in draft.

test/e2e/md_scale_test.go
pshail: We can extract out a func too for simplicity, the whole idea is not to miss the deferred error and this being a loop further complicates it

pshail: i still need to catch the deferred error though

pshail:
> 

early feedback is always great 👍

test/e2e/md_scale_test.go
pshail: Done with catching deferred error, still keeping it anon

test/e2e/e2e_test.go
pshail: i feel these checks should flag the error but not fail the test cases as these are house cleaning activities, but if we feel very strongly about erring as test failure then we will have to revisit these checks
- check if file exists [seen these being already removed before these house cleaning funcs are invoked and this introduces further flakiness if we enforce test failure on these failures
- clean/remove only if the files are not already removed

anusha94: Good point. These don't have to fail the test. I'm +1 on `clean/remove only if the files are not already removed`. I can open a new issue for this. wdyt?

anusha94 (approved):
lgtm

test/e2e/md_scale_test.go
dharmjit: Could we instead not use `defer` inside `for` loop. The improved version seems a bit hard to read.

dharmjit (commented):
LGTM, some nitpicks only.

agent/cloudinit/file_writer.go
```diff
@@ -62,7 +62,12 @@ func (w FileWriter) WriteToFile(file *Files) error {
 		return err
 	}
 
-	defer f.Close()
+	defer func() {
+		deferredErr := f.Close()
+		if deferredErr != nil {
+			fmt.Printf("Error closing file %s: %s", file.Path, deferredErr)
```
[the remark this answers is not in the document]
pshail: This is an error caught in defer function, AFAIK it won't allow the return of it, and in these particular cases didn't seem a candidate to panic and stop the world

agent/host_agent_suite_test.go
```diff
@@ -75,16 +73,20 @@ var _ = BeforeSuite(func() {
...
+	Expect(err).NotTo(HaveOccurred())
 	gexec.TerminateAndWait(time.Duration(10) * time.Second)
 	err = testEnv.Stop()
 	Expect(err).NotTo(HaveOccurred())
 })
 
 func writeKubeConfig() {
-	kubeconfigFile, err = ioutil.TempFile("", tmpFilePrefix)
+	kubeConf, err := ioutil.TempFile("", tmpFilePrefix)
+	kubeconfigFile = kubeConf
```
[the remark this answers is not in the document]
pshail: This is a weird one i agree, global variable shadowing is tricky, two options
- choose to live with that
- define an elaborate statefulness to return these data elements to be used across test steps or in end for cleanup/verification
- live with above weirdness

test/e2e/md_scale_test.go
pshail: the file is deferred close so that we can keep writing the docker logs till the function exits, if we close it just after the WriteDocker func call the purpose will be lost as that function does sync to agent log files in go routines AFAIR

agent/host_agent_suite_test.go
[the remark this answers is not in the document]
pshail: Fixed to some degree at-least in terms of code readability -> one place to set and rest of the places we get to read or clean the file
not too happy but better than before i guess

agent/installer/cli.go
```diff
@@ -88,15 +88,28 @@ func listSupported() {
...
-	defer w.Flush()
-
-	fmt.Fprintf(w, "%s\t%s\t%s\n", "OS", "K8S Version", "BYOH Bundle Name")
-	fmt.Fprintf(w, "%s\t%s\t%s\n", "---", "-----------", "----------------")
+	defer func() {
+		err := w.Flush()
+		if err != nil {
+			klogger.Error(err, "Failed to flush the tabwriter")
+		}
+	}()
```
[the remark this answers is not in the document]
pshail: AFAIK defer funcs can't return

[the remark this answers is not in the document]
pshail: TIL that they can :) will push a fix, there will be multiple fixes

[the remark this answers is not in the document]
pshail: Digging more i realised that none of the functions in cli return but handle error and the caller is not made aware of potential error, we should handle this in a new GH issue as this will change core logic wdyt?
For this fix i will keep it as in line with other steps in this function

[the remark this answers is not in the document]
pshail: @dharmjit ☝️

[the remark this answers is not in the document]
anusha94: Tldr - i think we can keep it as is

I wouldn't worry too much about cli.go right now because it's just used as testing installer changes locally. If we end up using it more or if the installer becomes complex, we can revisit then.

anusha94 (approved):

dharmjit (approved):

Issue #341, opened 2022-01-19
Improve codecov for agent/installer


anusha94:
At the moment, we have 69% coverage for installer. Identify more test cases to improve the coverage. To look at the detailed codecov report - https://app.codecov.io/gh/vmware-tanzu/cluster-api-provider-bringyourownhost

NilanjanDaw:
I can take a stab at this

NilanjanDaw:
`cli.go` is difficult to test and hard to mock. Recommending it to be added to the code-coverage report exclusion list [PR 344](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/issues/344#issuecomment-1026547102)
Some `bundle_downloader` paths need to be tested via integration tests through `reconciler` [PR 340](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/issues/340#issuecomment-1027681926).

Other paths have acceptable code coverage. Closing the issue.

PR #372, opened 2022-02-04
Files removed from test code coverage

### LINKED ISSUE #344: Ignore auto-generated files and files containing boilerplate code in codecov report
We have quite a bit of files that are auto-generated or contain just the boilerplate code which cannot really be unit/integration testable that can be captured in the codecov report.
This gives a false sense of underrated coverage which is not the exact indication of the overall value of  the repository. Codecov allows for ignoring such files and file paths. Refer https://docs.codecov.com/docs/ignoring-paths

Few files that can be ignored -
- https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/main/agent/main.go
- https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/main/controllers/infrastructure/byohost_controller.go
- https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/main/controllers/infrastructure/byomachinetemplate_controller.go

dharmjit:
- Auto-Generated files or files containing boilerplate code are removed from code coverage.
- Included webhook tests in code coverage

Fixes #344 

**Additional Information** :
`agent/main.go` is not removed as of now as some of the code related to flags is being tested

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -3,6 +3,11 @@ comment:                  # this is a top-level key
   behavior: default
   require_changes: true  # if true: only post the comment if coverage changes
 
+#list of paths (folders or file names) to exclude files from being collected by Codecov
 ignore:
 - "test"
-- "apis"
+- "controllers/infrastructure/byohost_controller.go"
+- "controllers/infrastructure/byomachinetemplate_controller.go"
+- "apis/infrastructure/v1beta1/*_types.go"
+- "apis/infrastructure/v1beta1/zz_generated.deepcopy.go"
+- "agent/installer/cli.go"
--- a/Makefile
+++ b/Makefile
@@ -55,10 +55,10 @@ all: build
 HOST_AGENT_DIR ?= agent
 
 # Run tests
-test: generate fmt vet manifests test-coverage webhook-test
+test: generate fmt vet manifests test-coverage
 
 test-coverage:
-	source ./scripts/fetch_ext_bins.sh; fetch_tools; setup_envs; ginkgo --randomizeAllSpecs -r --cover --coverprofile=cover.out --outputdir=. --skipPackage=test,apis .
+	source ./scripts/fetch_ext_bins.sh; fetch_tools; setup_envs; ginkgo --randomizeAllSpecs -r --cover --coverprofile=cover.out --outputdir=. --skipPackage=test .
 
 agent-test:
 	source ./scripts/fetch_ext_bins.sh; fetch_tools; setup_envs; ginkgo --randomizeAllSpecs -r $(HOST_AGENT_DIR) -coverprofile cover.out


anusha94 (approved):
awesome!!

codecov.yml
```diff
@@ -3,6 +3,11 @@ comment:                  # this is a top-level key
...
+#list of paths (folders or file names) to exclude files from being collected by Codecov
 ignore:
 - "test"
-- "apis"
\ No newline at end of file
+- "controllers/infrastructure/byohost_controller.go"
+- "controllers/infrastructure/byomachinetemplate_controller.go"
+- "apis/infrastructure/v1beta1/*_types.go"
+- "apis/infrastructure/v1beta1/zz_generated.deepcopy.go"
+- "agent/installer/cli.go"
```
jamiemonserrate, line 13: We have actual code in the `cli.go` right? This probably shouldn't be ignored?

dharmjit: This [Issue](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/issues/344) for improving agent/installer code test coverage was closed and It was found that `cli.go` is hard to test/mock[[here](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/issues/341#issuecomment-1027691657)] and hence we are excluding it from the code coverage.

@anusha94 @NilanjanDaw can add more.

pshail: lets open an issue to simplify code-design, wherever applicable for testability and address this overall not just cli.go
this PR can then proceed
@jamiemonserrate wdyt?

anusha94: Disclaimer - I am not saying no for an issue to simplify code-design.

We have 3 issues to cover existing code refactor / improvements / coverage - inputs on those issues / PRs would be helpful. See linked PRs in below issues
- https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/issues/221
- https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/issues/254
- https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/issues/340

With this, we hope to clear any tech debt we had accumulated at the time of v0.1.0 release. And this should reasonably improve the overall health and coverage. 
Coming to `cli.go` - this is used only to test the installer locally and is only an optional dev workflow. The functions that are called by `cli.go` are already covered by other unit / integration tests in the package. So this leaves us with writing tests for the flags supported by `cli.go`. We could most definitely do that, but didn't find much value add. And including this file in test coverage report will give a false indication of the overall health of the package.

So I had asked Nilanjan to mark this for exclusion. But if y'all think this needs to be tested for flags or any other use-case of `cli.go` I'm missing, happy to write tests. Then it makes sense to not exclude this file in codecov.

pshail: > Coming to cli.go - this is used only to test the installer locally and is only an optional dev workflow.
I guess then we should name it or put it in test package and call it out to show only a dev/test flow

+1 if the underlying code is already covered in other flows but the test coverage doesn't show that https://app.codecov.io/gh/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/main/agent/installer/cli.go

anusha94: > I guess then we should name it or put it in test package and call it out to show only a dev/test flow

Created https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/issues/374 to track this

pshail (approved):

PR #376, opened 2022-02-08
Fetch control plane IP at runtime

### LINKED ISSUE #206: Fetch the control plane endpoint ip programmatically for e2e tests
Currently, for the e2e tests, `CONTROL_PLANE_ENDPOINT_IP` is being passed as an environment variable in the [Makefile](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/main/Makefile#L112). It would be better to substitute the same by programmatically fetching it inside the tests.

anusha94:
Set the `CONTROL_PLANE_ENDPOINT_IP` programmatically at runtime during E2E tests based on docker network settings.

Fixes #206

### DIFF
--- a/test/e2e/e2e_docker_helper.go
+++ b/test/e2e/e2e_docker_helper.go
@@ -215,11 +215,18 @@ func setupByoDockerHost(ctx context.Context, clusterConName, byoHostName, namesp
 
 func setControlPlaneIP(ctx context.Context, dockerClient *client.Client) {
 	_, ok := os.LookupEnv("CONTROL_PLANE_ENDPOINT_IP")
-	if !ok {
-		inspect, _ := dockerClient.NetworkInspect(ctx, "kind", types.NetworkInspectOptions{})
-		ipOctets := strings.Split(inspect.IPAM.Config[0].Subnet, ".")
-		ipOctets[3] = "151"
-		ip := strings.Join(ipOctets, ".")
-		os.Setenv("CONTROL_PLANE_ENDPOINT_IP", ip)
+	if ok {
+		return
 	}
+	inspect, _ := dockerClient.NetworkInspect(ctx, "kind", types.NetworkInspectOptions{})
+	ipOctets := strings.Split(inspect.IPAM.Config[0].Subnet, ".")
+	
+	// The ControlPlaneEndpoint is a static IP that is in the hosts' 
+	// subnet but outside of its DHCP range. We believe 151 is a pretty 
+	// high number and we have < 10 containers being spun up, so we 
+	// can safely use this IP for the ControlPlaneEndpoint
+	ipOctets[3] = "151"
+	ip := strings.Join(ipOctets, ".")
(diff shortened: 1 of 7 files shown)


test/e2e/e2e_docker_helper.go
```diff
@@ -211,3 +212,14 @@ func setupByoDockerHost(ctx context.Context, clusterConName, byoHostName, namesp
 
 	return output, byohost.ID, err
 }
+
+func setControlPlaneIP(ctx context.Context, dockerClient *client.Client) {
```
anusha94, line 216: Add comment on what this function is doing. Because it is doing some specific hard coding to 151. So good to know why.
Something along the lines of
`The ControlPlaneEndpoint is a static IP that is in the hosts' subnet but outside of DHCP range. We believe 151 is a pretty high number and we have < 10 containers being spun up, so we can safely use this IP for the ControlPlaneEndpoint`

anusha94 (commented):
minor nits. Otherwise lgtm

test/e2e/e2e_docker_helper.go
NilanjanDaw: Added comments.

pshail:
small nit, rest is LGTM

NilanjanDaw:
> small nit, rest is LGTM

Hey, @pshail please Iet me know the edits required.

pshail:
@anusha94 just suggested the other way round  🙃  ... sorry for this @NilanjanDaw i just saw the resolved comments 
lets go ahead with this - its anyway a nit 😉

pshail (approved):

jamiemonserrate (approved):
LGTM

huchen2021:
I don't have strong opinion about this. I feel a little concern about this hard code way, it like a potential bug may occurs one day in the future. It will be better if  we can do this more smart. What I can think of is as followed, just for refer:
1) Fetch subNet by some method, for example, "docker network inspect kind | jq -r 'map(.IPAM.Config[].Subnet) []'"
2) Calculate the minIP and MaxIP.
3) Fetch all occupied ip list by  some method, for example, "docker network inspect kind | jq -r 'map(.Containers[].IPv4Address) []". 
4) Loop all IP between  minIP and MaxIP, and get a available one.

Anyway, if you are all fine by current fix, I am also ok with it.

anusha94:
>  I feel a little concern about this hard code way, it looks like a potential bug may occurs one day in the future.

I think that’s a very valid concern. We did think about it - but we know we won’t have that many containers now, so this felt a little complex to achieve something so small. Before this PR - we were hardcoding the full IP, [see this removed line from Makefile](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/commit/0fdb123357dc2a5457584ecd13f7893b7ab695bb#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L129), so compared to that, this PR added some smartness in fetching the subnet and hardcoding only the last octet.
We can revisit this in future if the `151` hardcoding is turning out to be problematic.

PR #385, opened 2022-02-15
More automation in getting_started.md guide

### LINKED ISSUE #272: More automation in getting_started.md guide
Can we improve the getting_started experience? At the moment, a developer would have to manually deploy a management cluster, build images, copy files to the docker container etc. 

Lets improve this experience so that we can delight the developer and have them able to try out our software with just one command run, and in under 10 mins.

huchen2021:
**What this PR does / why we need it:**
At the moment, a developer would have to manually deploy a management cluster, build images, copy files to the docker container etc.  Lets improve this experience so that we can delight the developer and have them able to try out our software with just one command run, and in under 10 mins.

Signed-off-by: chen hui <huchen@vmware.com>

Fixes #272

### DIFF
--- a/hack/getting_started.sh
+++ b/hack/getting_started.sh
@@ -297,11 +297,11 @@ function readArgs() {
                 shift 
                 ;;
             -m|--md) 
-                workerNums=$2
+                workerCount=$2
                 shift 2 
                 ;;
             -c|--cp) 
-                controlPlaneNums=$2
+                controlPlaneCount=$2
                 shift 2 
                 ;;
             -k|--kv) 
@@ -314,7 +314,7 @@ function readArgs() {
         esac
     done
 
-    byohNums=$[${workerNums}+${controlPlaneNums}]
+    byohNums=$[${workerCount}+${controlPlaneCount}]
 }
 
 function installCNI(){
@@ -478,7 +478,7 @@ function createWorkloadCluster() {
     # Find a available IP for control plane endpoint
     calcControlPlaneIP
 
-    CONTROL_PLANE_ENDPOINT_IP=${controlPlaneEndPointIp} clusterctl generate cluster ${workerClusterName} --infrastructure byoh --kubernetes-version ${kubernetesVersion} --control-plane-machine-count ${controlPlaneNums}  --worker-machine-count ${workerNums} --flavor docker > "${clusterYamlFile}"
+    CONTROL_PLANE_ENDPOINT_IP=${controlPlaneEndPointIp} clusterctl generate cluster ${workerClusterName} --infrastructure byoh --kubernetes-version ${kubernetesVersion} --control-plane-machine-count ${controlPlaneCount}  --worker-machine-count ${workerCount} --flavor docker > "${clusterYamlFile}"
     if [ $? -ne 0 ]; then
         echo "Generate ${clusterYamlFile} failed, exiting..."
         exit 1
@@ -529,8 +529,8 @@ byohImageTag="v1.22.3"
 managerClusterName="kind-byoh"
 workerClusterName="worker-byoh"
 controlPlaneEndPointIp=""
-workerNums=1
-controlPlaneNums=1
+workerCount=1
+controlPlaneCount=1
 byohNums=2
 defaultCni="0"
 manageClusterConfFile="${HOME}/.kube/management-cluster.conf"
--- a/hack/getting_started.sh
+++ b/hack/getting_started.sh
@@ -304,11 +304,12 @@ function createKindCluster(){
 }
 
 function cleanUp(){
+    local i=1
+
     kind delete clusters ${managerClusterName}
-    docker ps -a | grep host
-    for dockerId in `docker ps -a | grep host | awk '{print $1}'`
+    for ((i=1;i<=${byohNums};i++)); 
     do
-        docker rm -f ${dockerId}
+        docker rm -f host${i}
     done
 }
 
@@ -340,8 +341,6 @@ function readArgs() {
     done
 
     byohNums=$[${workerNums}+${controlPlaneNums}]
-
-    echo "defaultCni=${defaultCni}, controlPlaneNums=${controlPlaneNums}, workerNums=${workerNums}"
 }
 
 function installCNI(){
@@ -393,7 +392,6 @@ function retrieveKubeConfig() {
     exit
 }
 
-
 function checkNodeStatus() {
     local maxRunTimes=40
     local waitTime=30
@@ -433,21 +431,12 @@ function checkNodeStatus() {
 
 
 function prepareImageAndBinary() {
-    local  isImageExisted=0
-
     runCmd "cd ${reposDir}" 0
 
     # Build byoh image
     # Check if byoh image is existed
-    for line in `docker images | grep "${byohImageName}" | grep "${byohImageTag}" | awk '{print $1":"$2}'`
-    do
-        if [ ${line} == "${byohImageName}:${byohImageTag}" ]; then
-            isImageExisted=1
-            break
-        fi
-    done
-
-    if [ ${isImageExisted} -eq 0 ]; then
+    image=$(docker images ${byohImageName}:${byohImageTag} | grep -v REPOSITORY)
+    if [ -z "${image}" ]; then
         # The origin one will report error:   Could not connect to apt.kubernetes.io:443 (10.25.207.164), connection timed out [IP: 10.25.207.164 443]
         echo "deb http://packages.cloud.google.com/apt/ kubernetes-xenial main" > ${reposDir}/test/e2e/kubernetes.list
         runCmd "make prepare-byoh-docker-host-image" 0  "Making a byoh image: ${byohImageName}:${byohImageTag} ..."
@@ -465,11 +454,6 @@ function prepareImageAndBinary() {
     fi
 
     runCmd "cd -" 0
-
-
-    # Download byoh binary
-    #rm -f ${byohBinaryFile}
-    #runCmd "wget https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/releases/download/v0.1.0/byoh-hostagent-linux-amd64 -P /tmp" 0  "Downloading a byoh binary..."
 }
 
 function bringUpByoHost(){
@@ -543,6 +527,31 @@ function swapOff() {
 }
 
 
+function askForProceed() {
+    local warning='
+#####################################################################################################
+
+** WARNING **
+This modifys system settings - and do **NOT** revert them at the end of the test.
+
+It locally will change the following host config
+- disable swap, but it can revert back if rebooting vm
+- use "sudo apt-get update" command to download package information from all configured sources.
+- install docker, and enable it as service if not
+- install kind, clusterctl, jq, kubectl, build-essential and go, if not
+- create a kind cluster as manager cluster, byoh clustr as worker cluster
+
+#####################################################################################################'
+
+    echo "${warning}"
+	read -p "Do you want to proceed [Y/N]?" REPLY; 
+	if [[ ${REPLY} != "Y" && ${REPLY} != "y" ]]; then 
+        echo "Aborting..."
+        exit 1
+    fi
+}
+
+
 export PATH=/snap/bin:${PATH}
 byohImageName="byoh/node"
 byohImageTag="v1.22.3"
@@ -558,12 +567,12 @@ kubeConfigFile=/tmp/byoh-cluster-kubeconfig
 reposDir=$(dirname $0)/../
 byohBinaryFile=${reposDir}/bin/byoh-hostagent-linux-amd64
 
-
 if [ -z "${KUBERNETES_VERSION}" ]; then
     kubernetesVersion="v1.22.3"
 fi
 
 readArgs $@
+askForProceed
 swapOff
 intallDependencies 
 cleanUp
--- a/hack/getting_started.sh
+++ b/hack/getting_started.sh
@@ -1,19 +1,6 @@
 #!/bin/bash
-
-<<COMMENT
-ssh-copy-id <alias-name-for-this-jump-box>
-> sudo bash
-> chmod +w /etc/sudoers
-> vi /etc/sudoers
-<your-user-name>  ALL=(ALL) NOPASSWD:ALL
-> exit
-ssh-keygen -t rsa
-docker login
-
-COMMENT
-
 function isCmdInstalled() {
-    cmd=$1
+    local cmd=$1
     ${cmd} version 2>&1 | grep -q "not found"
     returnCode=$?  
     if [ ${returnCode} -eq 0 ] ; then
@@ -25,21 +12,21 @@ function isCmdInstalled() {
 }
 
 function exitIfNot(){
-    code=$1
-    expectedCode=$2
-    prompt=$3
+    local code=$1
+    local expectedCode=$2
+    local prompt=$3
     if [ ${code} == ${expectedCode} ]; then
         echo ${prompt}
-        exit
+        exit 1
     fi
 }
 
 function runCmd(){
-    cmd=$1
-    expectCode=$2
-    promptBeforeCmd=$3
-    PromptAfterCmdSuc=$4
-    promptAfterCmdFail=$5
+    local cmd=$1
+    local expectCode=$2
+    local promptBeforeCmd=$3
+    local PromptAfterCmdSuc=$4
+    local promptAfterCmdFail=$5
 
     if [ -n "${promptBeforeCmd}" ]; then
         echo ${promptBeforeCmd}
@@ -53,7 +40,7 @@ function runCmd(){
             echo ${promptAfterCmdFail}
         fi
         echo "${cmd} failed, exit...."
-        exit
+        exit 1
     fi
 
     if [ -n "${PromptAfterCmdSuc}" ]; then
@@ -61,64 +48,36 @@ function runCmd(){
     fi
 }
 
-function runCmdUntil() {
-    cmd=$1
-    expectCode=$2
-    waitTime=$3
-    maxRunTimes=$4
-    promptBeforeCmd=$5
-    PromptAfterCmdSuc=$6
-    promptAfterCmdFail=$7
-
-    echo "huchen: cmd is ${cmd}"
-
-    if [ -n "${promptBeforeCmd}" ]; then
-        echo ${promptBeforeCmd}
-    fi
-    for((i=1;i<=${maxRunTimes};i++));  
-    do   
-        ${cmd}
-        returnCode=$?
-        if [ ${returnCode} -ne ${expectCode} ]; then
-            sleep ${waitTime}
-            continue
-        else   
-            if [ -n "${PromptAfterCmdSuc}" ]; then
-                echo ${PromptAfterCmdSuc}
-            fi
-            return
-        fi
-        
-    done
-    if [ -n "${promptAfterCmdFail}" ]; then
-        echo ${promptAfterCmdFail}
-    fi
-    exit
-}
-
 function isIPOccupied() {
-    ip=$1
-    occupiedIPList=$2
+    local ip=$1
+    local occupiedIPList=$2
     echo "Checking if ${ip} is occupied"
 
-    echo "huchen: occupiedIPList is ${occupiedIPList}"
     IFS=$'\n'
     for occupiedIPWithMask in ${occupiedIPList}
     do
-        echo "huchen: occupiedIPWithMask is ${occupiedIPWithMask}"
         occupiedIP=`echo $occupiedIPWithMask | sed -En 's/^(.*)\/([0-9]{1,2})/\1/p'`
-        echo "huchen: ip is ${ip}, occupiedIP is ${occupiedIP}"
         if [ "${ip}" == "${occupiedIP}" ]; then
              return 1
         fi
     done
+
+    # should not use the gateway address
+    for gateway in ${gateways}
+    do 
+        if [ "${gateway}" == "${ip}" ]; then
+            return 1
+        fi
+    done
+
     echo "${ip} is not occupied"
     return 0
 }
 
 
+
 function num2IP() {
-    num=$1
+    local num=$1
     a=$((num>>24))
     b=$((num>>16&0xff))
     c=$((num>>8&0xff))
@@ -132,9 +91,9 @@ function binary2IP() {
     echo ${returnValue}
 }
 
-
 function calcControlPlaneIP() {
     subNet=`docker network inspect kind | jq -r 'map(.IPAM.Config[].Subnet) []'`
+    gateways=`docker network inspect kind | jq -r 'map(.IPAM.Config[].Gateway) []'`
     occupiedIPList=`docker network inspect kind | jq -r 'map(.Containers[].IPv4Address) []'`
 
     for line in ${subNet}
@@ -148,6 +107,7 @@ function calcControlPlaneIP() {
 
     echo "IPMask is ${subNet}"
     echo "occupiedIPList:${occupiedIPList}"
+    echo "gateways:${gateways}"
 
     ip=`echo $IPMask | sed -En 's/^(.*)\/([0-9]{1,2})/\1/p'`
     ipSubNetBit=`echo $IPMask | sed -En 's/^(.*)\/([0-9]{1,2})/\2/p'`
@@ -159,7 +119,7 @@ function calcControlPlaneIP() {
 
     ipSubNet=`echo ${binaryIP:0:${ipSubNetBit}}`
 
-    full0="00000000000000000000000000000010"
+    full0="00000000000000000000000000000001"
     full1="11111111111111111111111111111110"
 
     minIPBinary=${ipSubNet}`echo ${full0:${ipSubNetBit}}`
@@ -168,24 +128,30 @@ function calcControlPlaneIP() {
     minIPint=`echo "ibase=2; ${minIPBinary}"|bc`
     maxIPint=`echo "ibase=2; ${maxIPBinary}"|bc`
 
-
     for((i=${minIPint};i<=${maxIPint};i++));  
     do
         ip=$(num2IP ${i})
-
-        isIPOccupied "${ip}" "${occupiedIPList}"
+ 
+        isIPOccupied "${ip}" "${occupiedIPList}" "${gateways}"
         if [ $? == 0 ]; then
-            CONTROL_PLANE_ENDPOINT_IP=${ip}
-            echo "Available IP is ${CONTROL_PLANE_ENDPOINT_IP}"
+            controlPlaneEndPointIp=${ip}
+            echo "CONTROL_PLANE_ENDPOINT_IP is ${controlPlaneEndPointIp}"
             return
         fi
     done
+
+    echo "Can't get an available IP for control plane endpoint, exit...."
+    exit 1
+    
 }
 
-function configClusterctl() {
+function installByohProvider() {
+    local writeByoh=0
+    local clusterCtlYamlFile="${HOME}/.cluster-api/clusterctl.yaml"
+    local maxRunTimes=40
+    local waitTime=20
+
     # Write clusterctl.yaml
-    writeByoh=0
-    clusterCtlYamlFile="${HOME}/.cluster-api/clusterctl.yaml"
     if [ ! -f "${clusterCtlYamlFile}" ]; then
         writeByoh=1
         touch ${clusterCtlYamlFile}
@@ -208,44 +174,29 @@ EOF
     clusterctl config repositories | grep -q byoh
     if [ $? -ne 0 ] ; then
         echo "Config clusterctl failed..."
-        exit
+        exit 1
     fi
-}
-
-function runByohAgent(){
-    index=$1
-    byohBinaryFile=$2
-    manageClusterConfFile=$3
-
-    runCmd "docker cp ${byohBinaryFile} host${index}:/byoh-hostagent" 0 "Copying agent binary to byoh container: host${index}..."
-    runCmd "docker cp ${manageClusterConfFile} host${index}:/management-cluster.conf" 0 "Copying kubeconfig to byoh container: host${index}..."
-    docker exec -d host$i sh -c "chmod +x /byoh-hostagent && /byoh-hostagent --kubeconfig management-cluster.conf --skip-installation > /agent.log 2>&1"
-    if [ $? -ne 0 ] ; then
-        echo "Starting the host${index} agent..."
-        exit
-    fi
-
     
-    maxRunTimes=10
-    waitTime=1
+    runCmd "clusterctl init --infrastructure byoh" 0 " Transforming the Kubernetes cluster into a management cluster..."
+
+    # Waiting for byoh provider is totally ready
     for((i=1;i<=${maxRunTimes};i++));  
-    do   
-        kubectl get byohosts host${index} | grep -v NAME | grep -q host${index}
-        if [ $? -ne 0 ]; then
-            sleep ${waitTime}
-            continue
-        else   
-            echo "byohost object(host${index}) is created successfully..."
+    do  
+        byohStatus=$(kubectl get pods --all-namespaces | grep byoh-controller-manager | awk '{print $4}')
+        if [ "${byohStatus}" == "Running" ] ; then
+            echo "Byoh provider is ready"
             return
+        else
+            echo "Waiting for byoh-provider to be ready..."
+            sleep ${waitTime}
         fi
     done
-
-    echo "Error: byohost object(host${index}) is created failed..."
+    echo "Waiting too long for byoh provider, something may wrong with it."
     exit
 }
 
 function installDocker() {
-    cmdName="docker"
+    local cmdName="docker"
 
     ## check  if dependency is present
     isCmdInstalled "${cmdName}"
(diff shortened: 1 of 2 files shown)


scripts/getting_started.sh
```diff
@@ -0,0 +1,583 @@
...
+function installClusterctl() {
+
+    cmdName="clusterctl"
+     ## check  if denpency is installed before
+    isCmdInstalled "${cmdName}"
+
+    ## install denpency if it not installed
+    if [ $? -eq 0 ] ; then
+        echo "Installing ${cmdName}..."
+        curl -L https://github.com/kubernetes-sigs/cluster-api/releases/download/v1.0.0/clusterctl-linux-amd64 -o clusterctl && sudo install clusterctl /usr/local/bin/clusterctl
```
anusha94, line 338: we will need the latest version of clusterctl

scripts/getting_started.sh
```diff
@@ -0,0 +1,583 @@
...
+            exit
+        fi
+    else
+        echo "Create kind cluster successfully"
+    fi
+}
+
+function downloadByohCode(){
+    runCmd "rm -rf cluster-api-provider-bringyourownhost" 0 " Cleaning byoh code..."
+    msg=$(git clone git@github.com:vmware-tanzu/cluster-api-provider-bringyourownhost.git 2>&1)
```
anusha94, line 392: Why not just print the `git clone` output? We will not have to grep for error messages

scripts/getting_started.sh
```diff
@@ -0,0 +1,583 @@
...
+        if [ $? -ne 0 ] ; then
+            writeByoh=1
+        fi
+        grep -q "cert-manager" ${clusterCtlYamlFile}
+        if [ $? -ne 0 ] ; then
+            writeCertManager=1
+        fi
+    fi 
+
+    # if cert-manager missing,it report error: Error: action failed after 10 attempts: failed to get cert-manager object /, Kind=, /: Object 'Kind' is missing in 'unstructured object has no kind'
```
anusha94, line 198: You are probably getting this error because of an older version of clusterctl. This was fixed in `v1.0.4` or directly use the latest version of clusterctl

scripts/getting_started.sh
```diff
@@ -0,0 +1,583 @@
...
+    #runCmd "kind create cluster" 0 "Creating kind cluster..."
+    createKindCluster
+fi
+
+configClusterctl
+
+
+# check if init it before
+kubectl get pods --all-namespaces | grep -q byoh-controller-manager
+if [ $? -ne 0 ]; then
```
anusha94, line 427: Do we need to check this? What if we `init` even when provider is already installed? If it doesn't throw an error, we should be good, right?

scripts/getting_started.sh
```diff
@@ -0,0 +1,583 @@
...
+    runCmd "clusterctl init --infrastructure byoh" 0 " Transforming the Kubernetes cluster into a management cluster..."
+else
+    echo "clusterctl init --infrastructure byoh before"
+fi
+
+#check if byoh image is existed
+docker images | grep "byoh-dev/node" | grep -q "v1.22.3"
```
anusha94, line 434: same here. We should just run the make command to build the docker image

scripts/getting_started.sh
```diff
@@ -0,0 +1,583 @@
...
+byohBinaryFile="/tmp/byoh-hostagent-linux-amd64"
+if [ ! -f "${byohBinaryFile}" ]; then
+    runCmd "wget https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/releases/download/v0.1.0/byoh-hostagent-linux-amd64 -P /tmp" 0  "Downloading a byoh binary..."
+else
+    echo "${byohBinaryFile} existed."
+fi
+
+
+manageClusterConfFile="${HOME}/.kube/management-cluster.conf"
+if [ ! -f "${manageClusterConfFile}" ]; then
```
anusha94, line 458: same. Let's just copy and overwrite

scripts/getting_started.sh
```diff
@@ -0,0 +1,583 @@
...
+echo "Creating the workload cluster..."
+kubectl apply -f ${clusterYamlFile} 
+if [ $? -ne 0 ]; then
+    echo "Create the workload cluster failed"
+    exit
+fi
+
+echo "Retrieving the kubeconfig of workload cluster..."
+
+maxRunTimes=10
```
anusha94, line 528: why do we have to do this? Is this because we have to wait until the workload cluster is created?

scripts/getting_started.sh
```diff
@@ -0,0 +1,583 @@
...
+        continue
+    else 
+        kubectl get secret/byoh-cluster-kubeconfig -o json | jq -r .data.value | base64 --decode  > ${kubeConfigFile}
+        echo "Retrieve the kubeconfig of workload cluster successfully"
+        break
+    fi
+done
+
+
+echo "Applying a CNI for network..."
```
anusha94, line 545: We should allow for users to apply the CNI of their choice. I think it is better, we add a message that your cluster is provisioned, next step is to apply a CNI

scripts/getting_started.sh
huchen2021: That's because sometimes it doesn't have  correct access rights to code repos. I need to catch this error and print suggestion.

scripts/getting_started.sh
huchen2021: ok. Will try the latest one to see if it can fix this issue.

scripts/getting_started.sh
huchen2021: I found if we init again, it will report error and not continue this script, even caused some pod crash sometimes.
My script are designed that it can re-run again, again and again.

scripts/getting_started.sh
huchen2021: Yes, we need to wait until the workload cluster is ready.

huchen2021:
This is a script just do all steps in  getting_started.md guide, not designed to accepted any arguments. It just help customer try out our software with just one command run. If customer want to create "real" byoh culster with customize configuration, do it by manually. It's better not to rename this script name, or it will made customer think they can  bring up and configure their byoh cluster by this script.

> Suggest to move this script under `hack/` folder and change the name to `local-up.sh` Usually have seen a local-up script that can be used to bring up a minimal cluster for people to get started with

jamiemonserrate (commented):
Oh wow, this was a *lot* bigger than I expected. Thats a lot of code 😸 

Anyways, this feels like new functionality we are maintaining. I worry its going to be super easy to break this. So I suggest that we add some tests for it. Not unit tests, but at least an e2e test of sorts, that is included in our github checks. Essentially a check that runs this script, and verifies that we have a running cluster(s) at the end if it. Thoughts @huchen2021 ?

And, I assume this script would live in parallel to the existing BYOH getting started guide. Kind of like an alternate experience. If users want white glove service, they run this script. Else, users can get their hands dirty, and manually follow the steps, so that they know all the pieces that make up BYOH. So we would have to modify the README / exisiting getting started guide, to hook this in. Yes?

anusha94:
> Essentially a check that runs this script, and verifies that we have a running cluster(s) at the end if it

@jamiemonserrate, I don't know if that's necessary. This is sort of a simple script that will get a byoh workload cluster up. What breaking changes are you worried of? I am looking for a similar experience like k/k's [hack/local-up-cluster.sh](https://github.com/kubernetes/kubernetes/blob/master/hack/local-up-cluster.sh). 
The current script is hard coding way too many values - and this will become stale pretty soon. If we provide a way for the user to pass / set few args while also having some default values, we should be good (I _think_)

scripts/getting_started.sh
anusha94: @huchen2021 This is a good idea. @dharmjit & I have also thought about this in the past. Instead of having this as a release artifact, we should push this image to Harbor and we can pull it from there. This will be useful in e2e test as well.

huchen2021: @anusha94 Ok. I'll fix this after we have  this image to Harbor.

hack/getting_started.sh
```diff
@@ -0,0 +1,551 @@
+#!/bin/bash
+
+<<COMMENT
```
anusha94, line 3: Why do we need this?

hack/getting_started.sh
```diff
@@ -0,0 +1,551 @@
...
+function runCmdUntil() {
+    cmd=$1
+    expectCode=$2
+    waitTime=$3
+    maxRunTimes=$4
+    promptBeforeCmd=$5
+    PromptAfterCmdSuc=$6
+    promptAfterCmdFail=$7
+
+    echo "huchen: cmd is ${cmd}"
```
anusha94, line 73: Remove echo

hack/getting_started.sh
```diff
@@ -0,0 +1,551 @@
...
+    fi
+    exit
+}
+
+function isIPOccupied() {
+    ip=$1
+    occupiedIPList=$2
+    echo "Checking if ${ip} is occupied"
+
+    echo "huchen: occupiedIPList is ${occupiedIPList}"
```
anusha94, line 104: Remove echo

hack/getting_started.sh
```diff
@@ -0,0 +1,551 @@
...
+    IFS=$'\n'
+    for occupiedIPWithMask in ${occupiedIPList}
+    do
+        echo "huchen: occupiedIPWithMask is ${occupiedIPWithMask}"
```
anusha94, line 108: Please remove all custom echo messages that were used in debugging

hack/getting_started.sh
```diff
@@ -0,0 +1,551 @@
...
+
+    sudo systemctl status docker | grep -q "active (running)"
+    if [ $? -ne 0 ]; then
+        echo "Enable docker failed"
+        exit 
+    fi
+    echo "Enable docker success"
+
+    # Make sure current user has permission for docker, do this if not Create the docker group.
+    docker ps 2>&1 | grep -q "connect: permission denied"
```
anusha94, line 283: I think we should just exit with the error. This is making the script complicated and a getting started script should not be handling creating users / groups etc.

hack/getting_started.sh
```diff
@@ -0,0 +1,551 @@
...
+        ## check  if denpency is installed successfully
+        isCmdInstalled  "${cmdName}"
+        exitIfNot $? 0 "Installing ${cmdName} failed, exit..."
+    fi
+}
+
+function createKindCluster(){
+    echo  "Creating kind cluster..."
+
+    msg=$(kind create cluster 2>&1)
```
anusha94, line 362: Let's name this cluster as `kind-byoh`. Users might have a cluster called `kind` already present

hack/getting_started.sh
```diff
@@ -0,0 +1,551 @@
...
+        fi
+    else
+        echo "Download byoh code successfully"
+    fi
+}
+
+export PATH=/snap/bin:${PATH}
+intallDependencies 
+ 
+runCmd "sudo swapoff -a" 0 "Turning off swap..."
```
anusha94, line 393: Why should we turn off swap? Isn't the installer doing this for us?

hack/getting_started.sh
```diff
@@ -0,0 +1,551 @@
...
+else
+    echo "clusterctl init --infrastructure byoh before"
+fi
+
+#check if byoh image is existed
+docker images | grep "byoh-dev/node" | grep -q "v1.22.3"
+if [ $? -ne 0 ]; then
+    #runCmd "rm -rf cluster-api-provider-bringyourownhost" 0 " Cleaning byoh code..."
+    #runCmd "git clone git@github.com:vmware-tanzu/cluster-api-provider-bringyourownhost.git" 0 " Downloading byoh code..."
+    downloadByohCode
```
anusha94, line 422: Why do we have to download byoh code?
The user is running this script after cloning the byoh repo. Why do we need this clone again?

hack/getting_started.sh
anusha94: We should first check if `kind-byoh` is present. If so, delete and create a new one. 
Running this script means we are starting a fresh cluster. Should not rely on previous runs / states
This will save us lot of effort in checking if byoh is installed, cluster conf is present etc.

scripts/getting_started.sh
```diff
@@ -0,0 +1,583 @@
...
+CONTROL_PLANE_ENDPOINT_IP=""
+calcControlPlaneIP
+if [ -z "${CONTROL_PLANE_ENDPOINT_IP}" ]; then
+    echo "Can't get an available IP for control plane endpoint, exit...."
+    exit
+fi
+echo "CONTROL_PLANE_ENDPOINT_IP is ${CONTROL_PLANE_ENDPOINT_IP}"
+
+clusterYamlFile="/tmp/cluster-yaml"
+CONTROL_PLANE_ENDPOINT_IP=${CONTROL_PLANE_ENDPOINT_IP} clusterctl generate cluster byoh-cluster --infrastructure byoh --kubernetes-version v1.22.3 --control-plane-machine-count 1  --worker-machine-count 1 --flavor docker > "${clusterYamlFile}"
```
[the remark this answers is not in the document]
anusha94: I agree. But this is very much tightly coupled. Even if we want a different k8s version, we should change the script to support it.
We should take these values as args or env variables, but also define a default value. So that the user can use that if they wish different configurations.

scripts/getting_started.sh
anusha94: Again, I agree this being a getting-started script. But applying a CNI is out of scope of BYOH. What we claim is supporting different CNI, so our getting started script should not be biased to a particular CNI. 
Either give option on what cni user can install, or don't apply the CNI in the script - just print a message that byoh workload cluster is successfully created, next step is to apply a CNI of your choice.

hack/getting_started.sh
```diff
@@ -0,0 +1,551 @@
...
+}
+
+function runByohAgent(){
+    index=$1
+    byohBinaryFile=$2
+    manageClusterConfFile=$3
+
+    runCmd "docker cp ${byohBinaryFile} host${index}:/byoh-hostagent" 0 "Copying agent binary to byoh container: host${index}..."
+    runCmd "docker cp ${manageClusterConfFile} host${index}:/management-cluster.conf" 0 "Copying kubeconfig to byoh container: host${index}..."
+    docker exec -d host$i sh -c "chmod +x /byoh-hostagent && /byoh-hostagent --kubeconfig management-cluster.conf --skip-installation > /agent.log 2>&1"
```
anusha94, line 222: `--skip-installation` skips the installer part. We should not start the agent with this arg.

jamiemonserrate:
> What breaking changes are you worried of? I am looking for a similar experience like k/k's [hack/local-up-cluster.sh](https://github.com/kubernetes/kubernetes/blob/master/hack/local-up-cluster.sh)

I guess I'm worried that it would be super easy for this script to go out of date. If we introduce a breaking change, it won't be until much later that we find out. 

At the end of the day, its code. I am always a pessimist (things break), and prefer writing tests as much as possible. Happy to back down if you feel this test is too distracting or too much work to maintain.

hack/getting_started.sh
huchen2021: We use byoh-dev/node:v1.22.3 image, don't need install anything which is mentioned in getting_started.md.

anusha94: Looks like the guide needs to be updated. Please remove `--skip-installation` from the script and switch to non-dev image (the one that is used in e2e)

scripts/getting_started.sh
huchen2021: @anusha94 Installing CNI is also part of the content of getting_started.md. This script is just translate it to automatic script. And Installing CNI can also helps us to know better if our worker cluster is ok.

scripts/getting_started.sh
anusha94: @huchen2021 ,
Please try this image -
`docker pull projects.registry.vmware.com/cluster_api_provider_bringyourownhost/byoh/node-ubuntu-20.04:e2e`

scripts/getting_started.sh
anusha94: Yes, so this script also should be documenting in the Getting Started guide.
If a user wants to use the script, they can do that or do the steps manually.
CNI is a part of getting started guide, because it makes the cluster "complete". But it is after byoh has executed without any error, we apply the CNI.

The getting started script should provide a similar experience IMO. Or take an input from user for CNI of their choice, and then calico can be the default option.

anusha94:
@huchen2021 Since running this script will change the state on the host, we should display a warning message and take user's confirmation whether to proceed or not. We already do that in our e2e test -

https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/fb51a6d7684ebf99bba0260c346d315da4e8fe0c/Makefile#L140-L157

anusha94:
> I guess I'm worried that it would be super easy for this script to go out of date. If we introduce a breaking change, it won't be until much later that we find out.

@jamiemonserrate  😄 
Agreed. But it is there only to provide one-script experience of BYOH. If we keep few of the variables configurable and have sane default values, it won’t “break” as such. It will be stale, no doubt. We have to update this once in a while (and I think its okay and doesn’t have to be pointing to the latest artifacts all the time). We can do this as a one time ritual post a release, we anyway have to update our README and markdown files after every release. So it should be just about updating a few defaults.

hack/getting_started.sh
huchen2021: The installer run inside the container, and the container cannot manipulate its swap since it is inherited from the VM which it has no access to.

huchen2021:
> > I guess I'm worried that it would be super easy for this script to go out of date. If we introduce a breaking change, it won't be until much later that we find out.
> 
> @jamiemonserrate 😄 Agreed. But it is there only to provide one-script experience of BYOH. If we keep few of the variables configurable and have sane default values, it won’t “break” as such. It will be stale, no doubt. We have to update this once in a while (and I think its okay and doesn’t have to be pointing to the latest artifacts all the time). We can do this as a one time ritual post a release, we anyway have to update our README and markdown files after every release. So it should be just about updating a few defaults.

I totally agreed with @anusha94. We need to take care of this for every new release. It should be enough.

huchen2021:
> @huchen2021 Since running this script will change the state on the host, we should display a warning message and take user's confirmation whether to proceed or not. We already do that in our e2e test -
> 
> https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/fb51a6d7684ebf99bba0260c346d315da4e8fe0c/Makefile#L140-L157

Done. Thanks for your comment.

huchen2021:
@jamiemonserrate @anusha94

huchen2021:
This script will exit with code 1 if any error happened. It do checks on every step, and at the end of script, it also check if every node status is ok. Base on that, I add this test into github workflows, and it pass. You can know more detail by get-started-suite of [pull-390](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/pull/390).

 If we have to add a test for this script, @anusha94 @jamiemonserrate  What do you think about this way? If you are all think this is ok, I'll merge the test code part it into this PR.

hack/getting_started.sh
```diff
@@ -0,0 +1,592 @@
...
+
+}
+
+function installByohProvider() {
+    local writeByoh=0
+    local clusterCtlYamlFile="${HOME}/.cluster-api/clusterctl.yaml"
+    local maxRunTimes=40
+    local waitTime=20
+
+    # Write clusterctl.yaml
```
anusha94, line 154: Why is this logic to check if byoh already installed needs to be done? We can directly write the clusterctl.yaml and continue to install the provider. Since we are deleting kind-byoh cluster if it exists and creating a new one, it is safe to assume that this operation will not fail.

hack/getting_started.sh
```diff
@@ -0,0 +1,592 @@
...
+kubeConfigFile=/tmp/byoh-cluster-kubeconfig
+reposDir=$(dirname $0)/../
+byohBinaryFile=${reposDir}/bin/byoh-hostagent-linux-amd64
+
+if [ -z "${KUBERNETES_VERSION}" ]; then
+    kubernetesVersion="v1.22.3"
+fi
+
+readArgs $@
+askForProceed
```
anusha94, line 572:
```suggestion
userConfirmation
```

hack/getting_started.sh
```diff
@@ -0,0 +1,592 @@
...
+    kind create cluster --name ${managerClusterName}
+    if [ $? -ne 0 ]; then
+        echo "Create kind cluster failed"
+        exit 1
+    else
+        echo "Create kind cluster successfully"
+    fi
+}
+
+function cleanUp(){
```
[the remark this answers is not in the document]
huchen2021: No need to remove mgmt conf, it will be overwrite on every execution.

hack/getting_started.sh
```diff
@@ -0,0 +1,592 @@
...
+
+export PATH=/snap/bin:${PATH}
+byohImageName="byoh/node"
+byohImageTag="v1.22.3"
+managerClusterName="kind-byoh"
+workerClusterName="worker-byoh"
+controlPlaneEndPointIp=""
+workerNums=1
+controlPlaneNums=1
+byohNums=2
```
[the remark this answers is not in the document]
huchen2021: There are several place to used byohNums, it is better to calculate once. Not every time.
And For shell programming, numeric calculation is not as simple as string.

hack/getting_started.sh
huchen2021: The file used append mode, we can't directly write the clusterctl.yaml again. 
Are you sure that ${HOME}/.cluster-api/clusterctl.yaml is not possible to created by other?  If so, we could overwrite this file as a fresh one.

hack/getting_started.sh
```diff
@@ -0,0 +1,592 @@
...
+        echo "Apply a CNI for network failed"
+        exit 1
+    fi
+}
+
+function retrieveKubeConfig() {
+    local maxRunTimes=10
+    local waitTime=1
+
+    echo "Retrieving the kubeconfig of workload cluster..."
```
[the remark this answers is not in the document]
huchen2021: No, we need this one. After applying cluster-template, it need times to waiting for secret/${workerClusterName}-kubeconfig is ready. Trust, without this, the script is very unstable, I test it many many times.

hack/getting_started.sh
```diff
@@ -0,0 +1,592 @@
...
+workerNums=1
+controlPlaneNums=1
+byohNums=2
+defaultCni=0
+manageClusterConfFile="${HOME}/.kube/management-cluster.conf"
+kubeConfigFile=/tmp/byoh-cluster-kubeconfig
+reposDir=$(dirname $0)/../
+byohBinaryFile=${reposDir}/bin/byoh-hostagent-linux-amd64
+
+if [ -z "${KUBERNETES_VERSION}" ]; then
```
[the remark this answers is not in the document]
huchen2021: Sorry for that. I made misunderstanding about your previous comment. Will change this one.

huchen2021:
> > This script will exit with code 1 if any error happened. It do checks on every step, and at the end of script, it also check if every node status is ok. Base on that, I add this test into github workflows, and it pass. You can know more detail by get-started-suite of [pull-390](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/pull/390).
> > If we have to add a test for this script, @anusha94 @jamiemonserrate What do you think about this way? If you are all think this is ok, I'll merge the test code part it into this PR.
> 
> First of all, great job! Love that the script is running in ~8mins. But as we discussed earlier, I am leaning towards not having a test for this script. It is not expected to be at the latest state all the time and will require default updates with every release.

You are make a good point. But since this script only consume ~8mins, It wouldn't be introduce too much burden for github workflow. Anyway, I can accept both way.

@jamiemonserrate What do you think?

hack/getting_started.sh
```diff
@@ -0,0 +1,592 @@
+#!/bin/bash
+function isCmdInstalled() {
+    local cmd=$1
+    ${cmd} version 2>&1 | grep -q "not found"
```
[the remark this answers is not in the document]
huchen2021: I tried to remove version, but found issue with "jq 2>&1 | grep -q "not found"", it hangs.
Now all commands I used are work for "** version" right now. If there is command not work for version in the future, I'll do something in this function.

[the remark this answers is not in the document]
pshail: @huchen2021 @anusha94 what we observed here https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/main/agent/installer/checks.go#L14-L30 -> Potential of reuse if we extract a go lib out of it?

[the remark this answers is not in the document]
pshail: In the meantime check if "command -v" is a better alternative within this script base checker for a binary/command?
https://pubs.opengroup.org/onlinepubs/9699919799/utilities/command.html
```
$ command -V docker clusterctl
docker is /usr/local/bin/docker
clusterctl not found
```
or

```
$ command -v docker clusterctl
/usr/local/bin/docker
```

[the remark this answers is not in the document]
huchen2021: > @huchen2021 @anusha94 what we observed here https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/main/agent/installer/checks.go#L14-L30 -> Potential of reuse if we extract a go lib out of it?

How to reuse go function on shell script?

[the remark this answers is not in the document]
huchen2021: > In the meantime check if "command -v" is a better alternative within this script base checker for a binary/command? https://pubs.opengroup.org/onlinepubs/9699919799/utilities/command.html
> 
> ```
> $ command -V docker clusterctl
> docker is /usr/local/bin/docker
> clusterctl not found
> ```
> 
> or
> 
> ```
> $ command -v docker clusterctl
> /usr/local/bin/docker
> ```

command is a bash builtin which is not installed by default. It is not installed on customer's vm, it may needs to install it by "sudo apt-get install manpages-posix".  I am leaning towards not to have this one more dependency.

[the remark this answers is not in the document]
dharmjit: Suggestion: if we only wanted to see the tool is installed, Is `which` utility an option?

hack/getting_started.sh
dharmjit: Question: Do we even need to modify `clusterctl.yaml`, I believe byoh is already one of the supported provider in `clusterctl` via https://github.com/kubernetes-sigs/cluster-api/pull/5593

hack/getting_started.sh
```diff
@@ -0,0 +1,591 @@
...
+        echo "Config clusterctl failed..."
+        exit 1
+    fi
+
+    runCmd "clusterctl init --infrastructure byoh" 0 " Transforming the Kubernetes cluster into a management cluster..."
+
+    # Waiting for byoh provider is totally ready
+    for((i=1;i<=${maxRunTimes};i++));  
+    do  
+        byohStatus=$(kubectl get pods --all-namespaces | grep byoh-controller-manager | awk '{print $4}')
```
dharmjit, line 185: Suggestion: We could provide a specific namespace i.e. `byoh-system`. I was thinking maybe we could use deployment to see `byoh` provider is installed and running as below 
```
kubectl get deployment byoh-controller-manager -n byoh-system | awk '{print $4}'
```

huchen2021: It can easy to know the status of pods, if this is "Running", it means it is ok.
$ kubectl get pods --all-namespaces | grep byoh-controller-manager
byoh-system                         byoh-controller-manager-6b59775cfd-2bp92                        2/2     Running   0          26h

It'll be harder to extract the status of deployment, from the view of shell programming.
$ kubectl get deployment byoh-controller-manager -n byoh-system
NAME                      READY   UP-TO-DATE   AVAILABLE   AGE
byoh-controller-manager   1/1     1            1           26h

hack/getting_started.sh
anusha94: @dharmjit is that supported with CAPI `v1.0.4`?

dharmjit: This is introduced in https://github.com/kubernetes-sigs/cluster-api/releases/tag/v1.1.0 and I quickly verified this to be  working. Do we have to specifically use CAPI `v1.0.4`?
```
$ ./clusterctl-linux-amd64 init --infrastructure byoh
Fetching providers
Installing cert-manager Version="v1.5.3"
Waiting for cert-manager to be available...
Installing Provider="cluster-api" Version="v1.1.2" TargetNamespace="capi-system"
Installing Provider="bootstrap-kubeadm" Version="v1.1.2" TargetNamespace="capi-kubeadm-bootstrap-system"
Installing Provider="control-plane-kubeadm" Version="v1.1.2" TargetNamespace="capi-kubeadm-control-plane-system"
I0225 06:45:08.876085  569883 request.go:665] Waited for 1.035277754s due to client-side throttling, not priority and fairness, request: GET:https://127.0.0.1:37561/apis/clusterctl.cluster.x-k8s.io/v1alpha3?timeout=30s
Installing Provider="infrastructure-byoh" Version="v0.1.0" TargetNamespace="byoh-system"

Your management cluster has been initialized successfully!
```

hack/getting_started.sh
dharmjit: oh sorry, I wasn't making sense in my above comment. What I wanted to say instead is to validate on available > 0(running pods)

`kubectl get deployment byoh-controller-manager -n byoh-system --no-headers | awk '{print $4}'` 
or
`kubectl get deployment byoh-controller-manager -n byoh-system -o json | jq .status.readyReplicas`

These are just some alternatives and I am not suggesting one over the other.

hack/getting_started.sh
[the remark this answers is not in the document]
huchen2021: @dharmjit Sounds a good idea.

hack/getting_started.sh
huchen2021: @dharmjit Woo, you are right. This part followed getting_started.md, I think we need also to update this document.

anusha94: > Do we have to specifically use CAPI v1.0.4

I see your point now. We can use latest clusterctl in getting started script to avoid this block of code and directly use init, is that right? Makes perfect sense!
I was thinking to do that when rest of the codebase upgrades to latest version of CAPI, but I guess they need not be related.

hack/getting_started.sh
huchen2021: @dharmjit Got it, Thanks.

jamiemonserrate:
> > > This script will exit with code 1 if any error happened. It do checks on every step, and at the end of script, it also check if every node status is ok. Base on that, I add this test into github workflows, and it pass. You can know more detail by get-started-suite of [pull-390](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/pull/390).
> > > If we have to add a test for this script, @anusha94 @jamiemonserrate What do you think about this way? If you are all think this is ok, I'll merge the test code part it into this PR.
> > 
> > 
> > First of all, great job! Love that the script is running in ~8mins. But as we discussed earlier, I am leaning towards not having a test for this script. It is not expected to be at the latest state all the time and will require default updates with every release.
> 
> You are make a good point. But since this script only consume ~8mins, It wouldn't be introduce too much burden for github workflow. Anyway, I can accept both way.
> 
> @jamiemonserrate What do you think?

Happy to yield. Lets not test this with every PR then. 

PS - Thanks for taking the effort to get that workflow setup @huchen2021! Sorry, this was my bad you had to do extra work.

huchen2021:
> > > > This script will exit with code 1 if any error happened. It do checks on every step, and at the end of script, it also check if every node status is ok. Base on that, I add this test into github workflows, and it pass. You can know more detail by get-started-suite of [pull-390](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/pull/390).
> > > > If we have to add a test for this script, @anusha94 @jamiemonserrate What do you think about this way? If you are all think this is ok, I'll merge the test code part it into this PR.
> > > 
> > > 
> > > First of all, great job! Love that the script is running in ~8mins. But as we discussed earlier, I am leaning towards not having a test for this script. It is not expected to be at the latest state all the time and will require default updates with every release.
> > 
> > 
> > You are make a good point. But since this script only consume ~8mins, It wouldn't be introduce too much burden for github workflow. Anyway, I can accept both way.
> > @jamiemonserrate What do you think?
> 
> Happy to yield. Lets not test this with every PR then.
> 
> PS - Thanks for taking the effort to get that workflow setup @huchen2021! Sorry, this was my bad you had to do extra work.

@jamiemonserrate It's fine. I am very happy to do this. It's a study opportunity for me. Thanks for the responding.

hack/getting_started.sh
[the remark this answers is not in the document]
anusha94: yeah. This is fine I suppose 👍🏽

hack/getting_started.sh
```diff
@@ -0,0 +1,592 @@
...
+
+function prepareImageAndBinary() {
+    runCmd "cd ${reposDir}" 0
+
+    # Build byoh image
+    # Check if byoh image is existed
+    image=$(docker images ${byohImageName}:${byohImageTag} | grep -v REPOSITORY)
+    if [ -z "${image}" ]; then
+        # The origin one will report error:   Could not connect to apt.kubernetes.io:443 (10.25.207.164), connection timed out [IP: 10.25.207.164 443]
+        echo "deb http://packages.cloud.google.com/apt/ kubernetes-xenial main" > ${reposDir}/test/e2e/kubernetes.list
```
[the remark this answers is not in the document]
huchen2021: The origin one will report error:   Could not connect to apt.kubernetes.io:443 (10.25.207.164), connection timed out [IP: 10.25.207.164 443]
Had to do this to fix this error.

hack/getting_started.sh
```diff
@@ -0,0 +1,592 @@
...
+        runCmd "make prepare-byoh-docker-host-image" 0  "Making a byoh image: ${byohImageName}:${byohImageTag} ..."
+
+    else
+        echo "byoh image \"${byohImageName}:${byohImageTag}\" existed."
+    fi
+
+    # Build byoh binary
+    # Check if byoh binary is existed
+    if [ ! -f ${byohBinaryFile} ]; then
+        runCmd "make host-agent-binaries" 0  "Making byoh binary: ${byohBinaryFile} ..."
```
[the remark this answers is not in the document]
huchen2021: Because we build byoh image locally, and the version of host-agent-binaries should match this image. If not, it get errors sometimes.

anusha94 (approved):
LGTM.
Again, huge thank you for this PR 🎉

hack/getting_started.sh
```diff
@@ -0,0 +1,563 @@
+#!/bin/bash
+function isCmdInstalled() {
+    local cmd=$1
+    which ${cmd}
```
pshail, line 4: Recommend : command -v over which -> which is external to shell and is divergent when used across platforms, command is built-in
`$ which command
command: shell built-in command
$ command -v command
command
$ which docker
/usr/local/bin/docker
$ command -v docker
/usr/local/bin/docker
`
The above variability is *within* one platform in "which" output and hence not consistent

pshail:
Minor suggest rest LGTM

pshail (approved):

PR #388, opened 2022-02-16
Add unit tests for k8s install and uninstall components of agent reconciler

### LINKED ISSUE #340: Improve codecov for agent/reconciler

NilanjanDaw:
Add unit tests for the `installK8sComponents` and `uninstallk8sComponents` functions within reconcile.
However, since these functions deal with actual installation of the k8s components, the `Install()` and `Uninstall()`
methods are being mocked. A new interface `Installer` has been introduced to allow `counterfeiter` to mock the
methods.
 
Partial #340


agent/reconciler/host_reconciler.go
```diff
@@ -28,14 +29,26 @@ import (
...
 
+//go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 -generate
+
+//counterfeiter:generate . Installer
+type Installer interface {
+	Install(string, string, string, logr.Logger) error
+	Uninstall(string, string, string, logr.Logger) error
+}
+
+type K8sInstaller struct {
```
anusha94, line 40: this is not needed. We don't want to implement `Install` and `Uninstall` ourselves.

agent/main.go
```diff
@@ -159,8 +159,10 @@ func main() {
 			},
 		},
 		SkipInstallation: skipInstallation,
-		DownloadPath:     downloadpath,
 		Recorder:         mgr.GetEventRecorderFor("hostagent-controller"),
+		Installer: &reconciler.K8sInstaller{
```
anusha94, line 163: You should assign a reference of `installer.New` to Installer.

Instead of doing a `installer.New` in the `Install` & `Uninstall` functions you've defined below, it should be instantiated in `main.go` itself. Otherwise, we still have the same concern of instantiating installer multiple times and not actually taking the benefit of pre-checks.

agent/reconciler/host_reconciler.go
```diff
@@ -28,14 +29,26 @@ import (
...
+	DownloadPath string
+}
+
 type HostReconciler struct {
 	Client           client.Client
 	CmdRunner        cloudinit.ICmdRunner
 	FileWriter       cloudinit.IFileWriter
 	TemplateParser   cloudinit.ITemplateParser
 	Recorder         record.EventRecorder
+	Installer        Installer
```
anusha94, line 50:
```suggestion
	K8Installer        Installer
```
just to be more specific

agent/reconciler/host_reconciler.go
```diff
@@ -28,14 +29,26 @@ import (
 	infrastructurev1beta1 "github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/apis/infrastructure/v1beta1"
 )
 
+//go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 -generate
+
+//counterfeiter:generate . Installer
+type Installer interface {
+	Install(string, string, string, logr.Logger) error
+	Uninstall(string, string, string, logr.Logger) error
+}
```
dharmjit, line 38: We could rethink on creating an exportable `Installer Interface` in the `installer` package. There are not going to be multiple implementations but still it can provide value in testing and readability. thoughts?? @anusha94 @jamiemonserrate

anusha94: Yes, you're right. We need to rethink a few things on the installer package. But maybe not part of writing tests? These unit / integration tests have been deferred forever :D

anusha94: > provide value in testing and readability

True. The initial take was the installer will be exposed only via call to `New` and everything else should be accessed by the struct that is returned to facilitate data hiding (I think).

So maybe we don't want to change that behavior as part of this PR and take it up when we potentially add more features to the installer. Does that sound okay?

dharmjit: > So maybe we don't want to change that behavior as part of this PR and take it up when we potentially add more features to the installer. Does that sound okay?

yup definitely not in this PR, but just to keep in mind for future refactoring.

agent/main.go
```diff
@@ -159,8 +165,8 @@ func main() {
 			},
 		},
 		SkipInstallation: skipInstallation,
```
anusha94, line 167: suggestion - we can probably remove this field from `HostReconciler` struct. When doing `installer.New`, you can check for `--skip-installation` flag. If set, then `k8sInstaller` will be `nil`.
In the `Reconcile` function, instead of checking for `SkipInstallation` value, simply check on `k8sInstaller` to be `nil` or not

agent/reconciler/reconciler_test.go
```diff
@@ -41,14 +42,15 @@ var _ = Describe("Byohost Agent Tests", func() {
...
-			FileWriter:       fakeFileWriter,
-			TemplateParser:   fakeTemplateParser,
-			Recorder:         recorder,
-			SkipInstallation: true,
+			Client:         k8sClient,
+			CmdRunner:      fakeCommandRunner,
+			FileWriter:     fakeFileWriter,
+			TemplateParser: fakeTemplateParser,
+			Recorder:       recorder,
+			K8sInstaller:   nil,
```
anusha94, line 53: Shouldn't this be set to `fakeInstaller`?

NilanjanDaw: By default, all tests were written with `skipinstallation: true`, setting the `k8sInstaller: nil` will keep the same behavior.

agent/reconciler/reconciler_test.go
```diff
@@ -175,27 +177,28 @@ runCmd:
 					Expect(patchHelper.Patch(ctx, byoHost, patch.WithStatusObservedGeneration{})).NotTo(HaveOccurred())
 				})
 
-				// It("should set K8sComponentsInstallationSucceeded to false with Reason K8sComponentsInstallationFailedReason if Install fails", func() {
-				// 	fakeInstaller.InstallReturns(errors.New("k8s components install failed"))
+				It("should set K8sComponentsInstallationSucceeded to false with Reason K8sComponentsInstallationFailedReason if Install fails", func() {
+					hostReconciler.K8sInstaller = fakeInstaller
```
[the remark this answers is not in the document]
anusha94: I think ideally it should be the other way round. But this works for now. Let's proceed 👍🏽

anusha94 (approved):
Lgtm

agent/reconciler/reconciler_test.go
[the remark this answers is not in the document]
dharmjit: > I think ideally it should be the other way round. But this works for now. Let's proceed

@anusha94 This I guess is done when we reverted the changes last time.

dharmjit (approved):
lgtm

PR #399, opened 2022-02-24
Add host-agent integration tests

### LINKED ISSUE #393: Add integration tests for agent/reconciler
Add integration tests for agent/reconciler.
The current strategy is to run a single docker host and execute the `byoh-agent-binary` within it. We simulate `byoh machines`, `secrets` and everything else as required.
The integration tests will be focussed on the k8s components installer functions of the `host-agent`

NilanjanDaw:
Add integration tests for `installK8sComponents`. Also updates the existing `host-agent` tests to run within docker-based environments.

Fixes #393

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,21 +54,6 @@ all: build
 
 HOST_AGENT_DIR ?= agent
 
-# Run tests
-test: generate fmt vet manifests test-coverage
-
-test-coverage:
-	source ./scripts/fetch_ext_bins.sh; fetch_tools; setup_envs; ginkgo --randomizeAllSpecs -r --cover --coverprofile=cover.out --outputdir=. --skipPackage=test .
-
-agent-test:
-	source ./scripts/fetch_ext_bins.sh; fetch_tools; setup_envs; ginkgo --randomizeAllSpecs -r $(HOST_AGENT_DIR) -coverprofile cover.out
-
-controller-test:
-	source ./scripts/fetch_ext_bins.sh; fetch_tools; setup_envs; ginkgo --randomizeAllSpecs controllers/infrastructure -coverprofile cover.out
-
-webhook-test:
-	source ./scripts/fetch_ext_bins.sh; fetch_tools; setup_envs; ginkgo apis/infrastructure/v1beta1 -coverprofile cover.out
-
 ##@ General
 
 # The help target prints out all targets with their descriptions organized
@@ -124,6 +109,22 @@ prepare-byoh-docker-host-image:
 
 prepare-byoh-docker-host-image-dev:
 	docker build test/e2e -f docs/BYOHDockerFileDev -t ${BYOH_BASE_IMG_DEV}
+
+
+# Run tests
+test: generate fmt vet manifests test-coverage
+
+test-coverage:
+	source ./scripts/fetch_ext_bins.sh; fetch_tools; setup_envs; ginkgo --randomizeAllSpecs -r --cover --coverprofile=cover.out --outputdir=. --skipPackage=test .
+
+agent-test: docker-build prepare-byoh-docker-host-image 
+	source ./scripts/fetch_ext_bins.sh; fetch_tools; setup_envs; ginkgo --randomizeAllSpecs -r $(HOST_AGENT_DIR) -coverprofile cover.out
+
+controller-test:
+	source ./scripts/fetch_ext_bins.sh; fetch_tools; setup_envs; ginkgo --randomizeAllSpecs controllers/infrastructure -coverprofile cover.out
+
+webhook-test:
+	source ./scripts/fetch_ext_bins.sh; fetch_tools; setup_envs; ginkgo apis/infrastructure/v1beta1 -coverprofile cover.out
 
 test-e2e: take-user-input docker-build prepare-byoh-docker-host-image $(GINKGO) cluster-templates-e2e ## Run the end-to-end tests
 	$(GINKGO) -v -trace -tags=e2e -focus="$(GINKGO_FOCUS)" $(_SKIP_ARGS) -nodes=$(GINKGO_NODES) --noColor=$(GINKGO_NOCOLOR) $(GINKGO_ARGS) test/e2e -- \
--- a/Makefile
+++ b/Makefile
@@ -114,10 +114,10 @@ prepare-byoh-docker-host-image-dev:
 # Run tests
 test: generate fmt vet manifests test-coverage
 
-test-coverage: docker-build prepare-byoh-docker-host-image
+test-coverage: prepare-byoh-docker-host-image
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -125,6 +110,22 @@ prepare-byoh-docker-host-image:
 prepare-byoh-docker-host-image-dev:
 	docker build test/e2e -f docs/BYOHDockerFileDev -t ${BYOH_BASE_IMG_DEV}
 
+
+# Run tests
+test: generate fmt vet manifests test-coverage
+
+test-coverage: docker-build prepare-byoh-docker-host-image
```
anusha94, line 117: You don't need `docker-build` It is generating the controller manager image. For this test, probably just `prepare-byoh-docker-host-image` is enough

agent/host_agent_suite_test.go
```diff
@@ -110,3 +132,42 @@ func writeKubeConfig() {
 	_, err = getKubeConfig().Write(kubeConfigData)
 	Expect(err).NotTo(HaveOccurred())
 }
+
+func setupTestInfra(ctx context.Context, hostname, kubeconfig string, namespace *corev1.Namespace) *e2e.ByoHostRunner {
```
anusha94, line 136: Wondering if this is required. Can we initialize `byoHostRunner` in the `BeforeSuite` itself? Then things like `PathToHostAgentBinary`, `ByoHostName` and `Namespace` can be from inside each `BeforeEach`. wdyt?

agent/host_agent_test.go
```diff
@@ -71,53 +77,90 @@ var _ = Describe("Agent", func() {
...
+			Consistently(func() (done bool) {
+				_, err := os.Stat(agentLogFile)
+				if err == nil {
+					data, err := os.ReadFile(agentLogFile)
+					if err == nil && strings.Contains(string(data), "\"msg\"=\"error\"") {
+						return true
+					}
+				}
+				return false
+			}, 10).Should(BeFalse())
```
anusha94, line 101: Do we need `10s` here? Its not a time consuming operation on the agent side, right? The default timeout for `Consistently` should be enough I guess

agent/host_agent_test.go
```diff
@@ -71,53 +77,90 @@ var _ = Describe("Agent", func() {
...
+			Eventually(func() (done bool) {
+				_, err := os.Stat(agentLogFile)
+				if err == nil {
+					data, err := os.ReadFile(agentLogFile)
+					if err == nil && strings.Contains(string(data), "\"msg\"=\"error getting kubeconfig\"") {
+						return true
+					}
+				}
+				return false
+			}, 10).Should(BeTrue())
```
anusha94, line 127: Same as above. Do we need 10 here?

agent/host_agent_test.go
```diff
@@ -129,7 +172,7 @@ var _ = Describe("Agent", func() {
 					return nil
 				}
 				return createdByoHost
-			}).ShouldNot(BeNil())
+			}, 30).ShouldNot(BeNil())
```
anusha94, line 175: The registration should not take this long IMO. Have you tried without the extended timeout?

agent/host_agent_test.go
```diff
@@ -182,15 +225,96 @@ var _ = Describe("Agent", func() {
...
+				if err == nil {
+					data, err := os.ReadFile(agentLogFile)
+					if err == nil && strings.Contains(string(data), byoHost.Name) {
+						return true
+					}
+				}
+				return false
+			}, 10, 1).ShouldNot(BeTrue())
+		})
+		Context("when machineref & bootstrap is assigned", func() {
```
anusha94, line 256:
```suggestion
		Context("when machineref & bootstrap secret is assigned", func() {
```

agent/host_agent_suite_test.go
```diff
@@ -110,3 +132,42 @@ func writeKubeConfig() {
...
+	err = dockerClient.ContainerRemove(ctx, byoHostContainer.ID, dockertypes.ContainerRemoveOptions{})
+	Expect(err).NotTo(HaveOccurred())
+
+	err = k8sClient.Delete(ctx, namespace)
+	Expect(err).NotTo(HaveOccurred(), "failed to delete test namespace")
+
+	_, err = os.Stat(agentLogFile)
+	if err == nil {
+		err = os.Remove(agentLogFile)
+		if err != nil {
```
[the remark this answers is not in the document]
NilanjanDaw: Some tests do not write to the `agentLogFile`, the `os.Stat(agentLogFile)` check prevents us from erroneously trying to delete the log files in such a case.

dharmjit (commented):
lgtm, some nits only

agent/host_agent_suite_test.go
NilanjanDaw: I wondered about the same, but some of the `CommandArgs` passed during the tests are contradictory, and not properly clearing them up can lead to unpredictable behavior. For example, if the `--skip-installation` flag test executes before the k8s installation test and the `--skip-installation` flag is not cleared from the object, it can cause the k8s install test to fail.

dharmjit (approved):
lgtm!

agent/host_agent_test.go
```diff
@@ -190,7 +233,88 @@ var _ = Describe("Agent", func() {
...
+			var (
+				byoMachine *infrastructurev1beta1.ByoMachine
+				namespace  types.NamespacedName
+			)
+			BeforeEach(func() {
+				byoMachine = builder.ByoMachine(ns.Name, defaultByoMachineName).Build()
+				Expect(k8sClient.Create(ctx, byoMachine)).Should(Succeed())
+				byoHost := &infrastructurev1beta1.ByoHost{}
+				namespace = types.NamespacedName{Name: hostName, Namespace: ns.Name}
+				Eventually(func() (err error) {
```
anusha94, line 266: Can we replace this `Eventually` with this function call?
https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/6b69e60f40f964c03f429e0acf3e810804d5e268/controllers/infrastructure/suite_test.go#L146

anusha94: Although we might want to revisit writing / reading to cache. I don't think we read from cache anymore.
cc @dharmjit

anusha94 (approved):
lgtm.
Thank you for this PR!

agent/host_agent_test.go
NilanjanDaw: I think this requires a handle to the `reconciler`, which we do not have here.
https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/6b69e60f40f964c03f429e0acf3e810804d5e268/controllers/infrastructure/suite_test.go#L151

PR #400, opened 2022-02-25
Add support for host-agent integration tests

NilanjanDaw:
Add support functions required for integration tests to be added for the `host-agent`. The integration tests will reuse functions from the E2E tests with some refactors, which are also included here.

Partial #399 

**Additional information**
**Handle different network configurations for integration and E2E tests**
The E2E tests use `kind` for deploying the management cluster and then apply cluster-template in workload cluster. In the case of the integration tests, we wanted to use `envtest` instead, just like the controller integration tests.

The docker container creation helper methods were written with E2E tests in mind, so by default, they were creating containers in the `kind network`. However, in our case, the management cluster runs within the `host network`.

One workaround was to bind the `api-server` started by `envtest` to `0.0.0.0` so as to enable access from within Docker containers. However, the cert generated for the `api-server` by default is valid for localhost, and not for other addresses that we are binding to. One solution would have been to use `controller-runtime helpers` for generating certificates for the other addresses. However, we feel that would add unnecessary complexity to the tests.

Instead, we refactor the `createDockerContainer()` to dynamically attach the Docker container to the `host network` for envtest based clusters and to the `kind network` for E2E tests. This ensures that the management cluster and the hosts are in the same network.

**Handle redundant host-agent compilation**
The `SetupByoDockerHost()` was structured to always build the `host-agent` binary before setting up the docker containers. However, for tests involving multiple containers, this leads to the `host-agent` being compiled multiple times. To remove these redundant compilations, the agent build process has been moved to the BeforeEach() function of each test and a reference for the path to the compiled binary is passed to the `setupByoDockerHost()` function

**Separate Byoh docker create and exec**
The `SetupByoDockerHost()` has further been refactored to separate the docker creation and execution part, this would allow us to reuse the same docker container for multiple integration test cases

### DIFF
--- a/agent/host_agent_suite_test.go
+++ b/agent/host_agent_suite_test.go
@@ -23,6 +23,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/envtest"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/record"
+	bootstrapv1 "sigs.k8s.io/cluster-api/bootstrap/kubeadm/api/v1beta1"
+	ctrl "sigs.k8s.io/controller-runtime"
 )
 
 var (
@@ -31,6 +36,21 @@ var (
 	k8sClient             client.Client
 	tmpFilePrefix         = "kubeconfigFile-"
 	testEnv               *envtest.Environment
+
+	clientFake client.Client
+	// byoClusterReconciler  *controllers.ByoClusterReconciler
+	byoCluster            *infrastructurev1beta1.ByoCluster
+	capiCluster           *clusterv1.Cluster
+	defaultClusterName    = "my-cluster"
+	defaultNodeName       = "my-host"
+	defaultByoHostName    = "my-host"
+	defaultMachineName    = "my-machine"
+	defaultByoMachineName = "my-byomachine"
+	defaultNamespace      = "default"
+	fakeBootstrapSecret   = "fakeBootstrapSecret"
+	recorder              *record.FakeRecorder
+	k8sManager            ctrl.Manager
+	cfg                   *rest.Config
 )
 
 func TestHostAgent(t *testing.T) {
@@ -52,8 +72,13 @@ var _ = BeforeSuite(func() {
 		CRDDirectoryPaths: []string{
 			filepath.Join("..", "config", "crd", "bases"),
 			filepath.Join(build.Default.GOPATH, "pkg", "mod", "sigs.k8s.io", "cluster-api@v1.0.4", "config", "crd", "bases"),
+			filepath.Join(build.Default.GOPATH, "pkg", "mod", "sigs.k8s.io", "cluster-api@v1.0.4", "bootstrap", "kubeadm", "config", "crd", "bases"),
 		},
+
 		ErrorIfCRDPathMissing: true,
+		KubeAPIServerFlags: []string{
+			"--advertise-address=10.148.66.54",
+		},
 	}
 
 	cfg, err := testEnv.Start()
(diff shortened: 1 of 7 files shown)


agent/host_agent_suite_test.go
```diff
@@ -52,7 +52,9 @@ var _ = BeforeSuite(func() {
 		CRDDirectoryPaths: []string{
 			filepath.Join("..", "config", "crd", "bases"),
 			filepath.Join(build.Default.GOPATH, "pkg", "mod", "sigs.k8s.io", "cluster-api@v1.0.4", "config", "crd", "bases"),
+			filepath.Join(build.Default.GOPATH, "pkg", "mod", "sigs.k8s.io", "cluster-api@v1.0.4", "bootstrap", "kubeadm", "config", "crd", "bases"),
```
anusha94, line 55: this is not required at the moment

agent/host_agent_test.go
```diff
@@ -12,6 +12,7 @@ import (
 	"os"
 	"os/exec"
 	"runtime"
+
```
anusha94, line 15: How about refactoring existing tests in this file to be run inside the container? That we can validate the helper function changes in this PR.

test/e2e/e2e_docker_helper.go
```diff
@@ -213,6 +231,58 @@ func setupByoDockerHost(ctx context.Context, clusterConName, byoHostName, namesp
 	return output, byohost.ID, err
 }
 
+func SetupByoDockerHostWithConfig(ctx context.Context, byoHostName, port, namespace string, dockerClient *client.Client, kubeconfigFile *os.File, flags map[string]string) (types.HijackedResponse, string, error) {
```
anusha94, line 234: This and SetupByoDockerHost is almost duplicate code. Can we try and make it as a single func with different configurations? 
The function arguments are exploding, so good time to define a struct?

On a side note - this file need not be called e2e_docker_helper anymore. Maybe just docker_helper will do.

test/e2e/docker_helper.go
```diff
@@ -142,74 +155,97 @@ func copyToContainer(ctx context.Context, cli *client.Client, copyConfig cpConfi
...
+		Expect(len(containers)).To(Equal(1))
+
+		profile, err := r.DockerClient.ContainerInspect(r.Context, containers[0].ID)
+		Expect(err).NotTo(HaveOccurred())
+
+		kubeconfig, err = os.ReadFile(r.bootstrapClusterProxy.GetKubeconfigPath())
+		Expect(err).NotTo(HaveOccurred())
+
+		re := regexp.MustCompile("server:.*")
+		kubeconfig = re.ReplaceAll(kubeconfig, []byte("server: https://"+profile.NetworkSettings.Networks["kind"].IPAddress+":6443"))
```
anusha94, line 201:
```suggestion
		kubeconfig = re.ReplaceAll(kubeconfig, []byte("server: https://"+profile.NetworkSettings.Networks[r.NetworkInterface].IPAddress+":6443"))
```

anusha94 (approved):
This looks good. We can do more refactoring when adding new tests. Since all the checks are passing, and we are unblocked on refactored docker helper methods, let's get this merged and then iterate!

test/e2e/docker_helper.go
```diff
@@ -142,74 +155,97 @@ func copyToContainer(ctx context.Context, cli *client.Client, copyConfig cpConfi
...
+		listopt.Filters.Add("name", r.ByoHostName)
+		containers, err := r.DockerClient.ContainerList(r.Context, listopt)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(len(containers)).To(Equal(1))
+
+		kubeconfig, err = os.ReadFile(r.KubeconfigFile.Name())
+		Expect(err).NotTo(HaveOccurred())
+
+		re := regexp.MustCompile("server:.*")
+		kubeconfig = re.ReplaceAll(kubeconfig, []byte("server: https://127.0.0.1:"+r.Port))
```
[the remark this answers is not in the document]
NilanjanDaw: We need to update the port number, which is ephemeral.

PR #416, opened 2022-03-02
build byoh binary only once to save test time

huchen2021:
It takes time to building host agent separately for every container. If we move it to suit-test.go to run it only once for entire e2e. It can save some time. My test is as followed:

Before do this:
-     e2e_test: time elapse: 4m38.405165171s
-     md_scale_test: time elapse: 11m2.787283875s
-     byohost_reuse_test: time elapse: 7m14.318728577s

After do this:
-     e2e_test: time elapse: 5m27.469585284s
-     md_scale_test: time elapse: 7m26.022867913s
-     byohost_reuse_test: time elapse: 6m51.386107789s

It can save more time for md_scale_test obviously.

Plus, I found some issue our code, fix it in meantime.
-     Writing byoh docker logs is not working, that's because we close file too early.
-     Should not let md_scale_test and byohost_reuse_test depend on the code of e2e_test.

Signed-off-by: chen hui <huchen@vmware.com>

### DIFF
--- a/test/e2e/e2e_suite_test.go
+++ b/test/e2e/e2e_suite_test.go
@@ -106,7 +106,7 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 	initBootstrapCluster(bootstrapClusterProxy, e2eConfig, clusterctlConfigPath, artifactFolder)
 
 	var err error
-	By("huchen: Start to build host agent binary")
+	By("building host agent binary")
 	pathToHostAgentBinary, err = gexec.Build("github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/agent")
 	Expect(err).NotTo(HaveOccurred())
 
--- a/test/e2e/e2e_suite_test.go
+++ b/test/e2e/e2e_suite_test.go
@@ -15,6 +15,7 @@ import (
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/gexec"
 	infraproviderv1 "github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/apis/infrastructure/v1beta1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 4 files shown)


test/e2e/md_scale_test.go
```diff
@@ -94,14 +87,13 @@ var _ = Describe("When testing MachineDeployment scale out/in", func() {
...
 			// read the log of host agent container in backend, and write it
 			agentLogFile := fmt.Sprintf("/tmp/host-agent-%d.log", i)
-			func() {
-				f := WriteDockerLog(output, agentLogFile)
-				defer func() {
-					deferredErr := f.Close()
-					if deferredErr != nil {
-						Showf("error closing file %s:, %v", agentLogFile, deferredErr)
-					}
-				}()
```
dharmjit, line 104: This change is done as part of https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/pull/358 and the idea is to capture deferred errors inside a loop.

test/e2e/e2e_suite_test.go
```diff
@@ -95,6 +105,11 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 	By("Initializing the bootstrap cluster")
 	initBootstrapCluster(bootstrapClusterProxy, e2eConfig, clusterctlConfigPath, artifactFolder)
 
+	var err error
+	By("huchen: Start to build host agent binary")
```
dharmjit, line 109:
```suggestion
	By("building host agent binary")
```

dharmjit (commented):
lgtm, basic nits only.

test/e2e/md_scale_test.go
huchen2021: The defer function is executed when leaving func at line 97, which is not right. It caused no agent_log generated for md_scale_test.

dharmjit: > It caused no agent_log generated for md_scale_test

Could you elaborate a bit on this? Do you mean that this defer execution is required at the end of the `It` test ?
cc: @pshail

huchen2021: Yes, Because there is a thread running backend to write docker log to file, we should close the file at the end of test.

dharmjit (approved):
lgtm

PR #429, opened 2022-03-10
Add feature gate support for SecureAccess to host agent

### LINKED ISSUE #428: Enable feature gate for secure access
Add a feature gate flag called `SecureAccess` to enable bootstrap-token based cluster access for BYOH

NilanjanDaw:
Adds feature-gate support for `SecureAccess` to host agent. Additionally,
- Offloads the agent registration process to the proposed `ByoAdmission Controller` when the `SecureAccess` flag is active.
- Updates the `hostReconciler` to delay network interface setup until host registration.

Fixes #428

### DIFF
--- a/agent/main.go
+++ b/agent/main.go
@@ -10,14 +10,15 @@ import (
 	"os"
 	"strings"
 
-	"github.com/spf13/pflag"
+	"github.com/go-logr/logr"
+	pflag "github.com/spf13/pflag"
 	"github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/agent/cloudinit"
-	"github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/agent/feature"
 	"github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/agent/installer"
 	"github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/agent/reconciler"
 	"github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/agent/registration"
 	"github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/agent/version"
 	infrastructurev1beta1 "github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/apis/infrastructure/v1beta1"
+	"github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/feature"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -94,6 +95,17 @@ func setupflags() {
 	feature.MutableGates.AddFlag(pflag.CommandLine)
 }
 
+func handleHostRegistration(k8sClient client.Client, hostName string, logger logr.Logger) (err error) {
+	if feature.Gates.Enabled(feature.SecureAccess) {
+		logger.Info("secure access enabled, waiting for host to be registered by ByoAdmission Controller")
+	} else {
+		registration.LocalHostRegistrar = &registration.HostRegistrar{K8sClient: k8sClient}
+		err := registration.LocalHostRegistrar.Register(hostName, namespace, labels)
+		return err
+	}
+	return nil
+}
+
 var (
 	namespace          string
 	scheme             *runtime.Scheme
@@ -141,8 +153,7 @@ func main() {
 		return
 	}
 
-	registration.LocalHostRegistrar = &registration.HostRegistrar{K8sClient: k8sClient}
-	err = registration.LocalHostRegistrar.Register(hostName, namespace, labels)
+	err = handleHostRegistration(k8sClient, hostName, logger)
 	if err != nil {
 		logger.Error(err, "error registering host %s registration in namespace %s", hostName, namespace)
 		return
@@ -190,10 +201,6 @@ func main() {
 		},
 		Recorder:     mgr.GetEventRecorderFor("hostagent-controller"),
 		K8sInstaller: k8sInstaller,
-	}
-
-	if feature.Gates.Enabled(feature.SecureAccess) {
-		logger.Info("secure access enabled")
 	}
 
 	if err = hostReconciler.SetupWithManager(context.TODO(), mgr); err != nil {
--- a/agent/main.go
+++ b/agent/main.go
@@ -106,6 +106,21 @@ func handleHostRegistration(k8sClient client.Client, hostName string, logger log
 	return nil
 }
 
+func setupTemplateParser() *cloudinit.TemplateParser {
+	var templateParser *cloudinit.TemplateParser
+	if feature.Gates.Enabled(feature.SecureAccess) {
+		templateParser = nil
+	} else {
+		templateParser = &cloudinit.TemplateParser{
+			Template: registration.HostInfo{
+				DefaultNetworkInterfaceName: registration.LocalHostRegistrar.ByoHostInfo.DefaultNetworkInterfaceName,
+			},
+		}
+	}
+
+	return templateParser
+}
+
 var (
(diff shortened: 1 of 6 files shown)


NilanjanDaw:
One of the tasks for this PR is to prevent `host registration` if `SecureAccess` is enabled and instead depend on the `ByoAdmission Controller` to perform the registration. However, part of the current registration process also involves setting up the network interfaces:
https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/cc9ae9bda980f2a119b494e4caf518b475151186/agent/registration/host_registrar.go#L67

This in turn is required in the`bootstrapK8sNode` process as part of `reconcileNormal`:
https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/cc9ae9bda980f2a119b494e4caf518b475151186/agent/main.go#L180-L191

https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/cc9ae9bda980f2a119b494e4caf518b475151186/agent/reconciler/host_reconciler.go#L247-L254

Should we refactor the registration process and separate the network interface setup? @anusha94 @dharmjit wdyt?
cc @sachinkumarsingh092

feature/feature.go
```diff
@@ -0,0 +1,28 @@
...
+	Gates        featuregate.FeatureGate        = MutableGates
+)
+
+func init() {
+	runtime.Must(MutableGates.Add(defaultClusterAPIFeatureGates))
+}
+
+// defaultClusterAPIFeatureGates consists of all known cluster-api-specific feature keys.
+// To add a new feature, define a key for it above and add it here.
+var defaultClusterAPIFeatureGates = map[featuregate.Feature]featuregate.FeatureSpec{
```
anusha94, line 26:
```suggestion
var defaultClusterAPIBYOHFeatureGates = map[featuregate.Feature]featuregate.FeatureSpec{
```

feature/feature.go
```diff
@@ -0,0 +1,28 @@
+// Copyright 2021 VMware, Inc. All Rights Reserved.
```
anusha94, line 1:
```suggestion
// Copyright 2022 VMware, Inc. All Rights Reserved.
```

agent/main.go
```diff
@@ -90,6 +92,33 @@ func setupflags() {
 	for _, hiddenFlag := range hiddenFlags {
 		_ = pflag.CommandLine.MarkHidden(hiddenFlag)
 	}
+	feature.MutableGates.AddFlag(pflag.CommandLine)
+}
+
+func handleHostRegistration(k8sClient client.Client, hostName string, logger logr.Logger) (err error) {
+	if feature.Gates.Enabled(feature.SecureAccess) {
```
anusha94, line 99: what about setting as an env variable? Something like `SECURE_ACCESS=true`

NilanjanDaw: So in the case of feature-gate implementations for CAPI, I think they're just using the environment variables to set the flag, something like this:
`--feature-gates=ClusterTopology=${CLUSTER_TOPOLOGY:=false},KubeadmBootstrapFormatIgnition=${EXP_KUBEADM_BOOTSTRAP_FORMAT_IGNITION:=false}` Source: [Cluster Topology feature flag](https://github.com/kubernetes-sigs/cluster-api/blob/5756be13affef606b6683c2c413f0987c1f0e451/controlplane/kubeadm/config/manager/manager.yaml#L24)

I don't think we need to do anything in the code to handle this.

agent/main.go
```diff
@@ -90,6 +92,33 @@ func setupflags() {
...
+		templateParser = nil
+	} else {
+		templateParser = &cloudinit.TemplateParser{
+			Template: registration.HostInfo{
+				DefaultNetworkInterfaceName: registration.LocalHostRegistrar.ByoHostInfo.DefaultNetworkInterfaceName,
+			},
+		}
+	}
+
+	return templateParser
```
dharmjit, line 121: Question: `HostReconciler` will get called only in case of some Create/Update/Delete event of `ByoHost`. I am not sure if we require some handling for template parser.

NilanjanDaw: The `templateParser` is part of the `HostReconciler` bootstrap process which in turn is used to bootstrap the node. The `setupTemplateParser` sets up a minimal `reconciler` which will get updated once the `byohost` is registered.

dharmjit: > The setupTemplateParser sets up a minimal reconciler which will get updated once the byohost is registered

What's the limitation if we create the `HostReconciler` the same way as of today?

NilanjanDaw: As discussed [here](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/issues/428#issuecomment-1067645100), the `HostReconciler` has some dependencies like network interface setup, with the host registration process. Since we are skipping the registration, the `HostReconciler` will have to be updated.

agent/main.go
anusha94: That is if the env variables are passed as args via the `manager.yaml`
For host agent, we don't have anything like the `manager.yaml` and we start the binary directly.

What would happen if the user set the env variable and installed the byoh provider? 
1. the host agent is unaware of any such variable (the controllers on mgmt cluster would enable SecureAccess and agent controllers wouldn't know about it)
2. these variables need to be set on the host itself

@dharmjit Thinking about it, setting env variable on each host is tedious. Maybe the info should flow from the cluster template. But for this PR, let's assume the env variable needs to be set on the host so that we don't do any api / controller change for now.

Thoughts? @NilanjanDaw

anusha94: Okay realising we cannot rely on info from cluster template. This step is before applying the cluster template.

NilanjanDaw: Yeah, this makes sense. The host-agent requirement is unique in this regard compared to the ones in CAPI.

agent/main.go
```diff
@@ -139,8 +168,7 @@ func main() {
 		return
 	}
 
-	registration.LocalHostRegistrar = &registration.HostRegistrar{K8sClient: k8sClient}
-	err = registration.LocalHostRegistrar.Register(hostName, namespace, labels)
+	err = handleHostRegistration(k8sClient, hostName, logger)
```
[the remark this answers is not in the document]
NilanjanDaw: I am wondering if initializing the `registration.LocalHostRegistrar` regardless of the `SecureAccess` flag status will have any unintended consequences.

agent/main.go
```diff
@@ -90,6 +92,33 @@ func setupflags() {
...
+		registration.LocalHostRegistrar = &registration.HostRegistrar{K8sClient: k8sClient}
+		err := registration.LocalHostRegistrar.Register(hostName, namespace, labels)
+		return err
+	}
+	return nil
+}
+
+func setupTemplateParser() *cloudinit.TemplateParser {
+	var templateParser *cloudinit.TemplateParser
+	if feature.Gates.Enabled(feature.SecureAccess) {
```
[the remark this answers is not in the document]
NilanjanDaw: Sure. Curiosity question, what is the advantage of one over the other?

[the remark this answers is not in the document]
anusha94: Not really. But setting up template parser doesn't care about feature flag. Setting the feature flag causes registrar to be nil and because registrar is nil, you cannot set template parser field.

If suppose you have to add more conditionals around registrar, you will have to repeat that condition here. 
But this is a minor suggestion though. Feel free to disagree :)

anusha94:
lgtm

anusha94:
lgtm

agent/main.go
NilanjanDaw: As discussed, we will hold off on adding support for environment variables in a future PR.

anusha94 (approved):
lgtm.

dharmjit (approved):
lgtm

PR #430, opened 2022-03-10
Implemented K8s patch version masking in the registry matrix of the Installer

### LINKED ISSUE #411: [Feature Enhancement] Make installer flexible to support any entire K8s minor series version
Currently, the agent installer is tightly coupled with K8s version `v1.22.3`. This means for any other k8s version support, even a different patch of the same `v1.22` series will not be supported by the installer. For a user, it means that whatever host-agent binary they downloaded for a given version, will not be applicable for a future patch version of the same k8s minor series. The host-agent will have to be upgraded in such scenarios.

This is not a very good user experience, and also not scalable to add support for every new patch of k8s release. Instead, we should make the installer flexible enough to be supporting a particular minor series - like `v1.22.*`. Now any patch version in this series will work out of the box - given that the corresponding bundle is present in the OCI registry.

Few pointers in the code base that might help when working on the PR for this enhancement -
Remove hard coding of a particular patch version and introduce something like `v1.22.*` -
https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/67b626e610dc9721c397dad7afdc10eef8758503/agent/installer/installer.go#L50-L65

**Goals**
- Remove existing hard coding of v1.22.3 support
- Support v1.22.* K8s version.
- We already have v1.22.3 bundle in the OCI repo (```docker pull projects.registry.vmware.com/cluster_api_provider_bringyourownhost/byoh-bundle-ubuntu_20.04.1_x86-64_k8s_v1.22.3:
v0.1.0_alpha.2```), add a new bundle to support `v1.22.5` which can used for testing this issue

**Non Goals**
- Support any other k8s minor series other than `v1.22`

**Note**
We need to revisit the tagging scheme while pushing the bundle to the repo (We think release tag of byoh is not a good indicator). Please contact the maintainers to help push the bundle to the registry.

jeuxdemains:
Implements masking of patch versions of K8s (e.g.: 1.22.*)

Fixes #411

**Additional information**
The way it works is the same as the OS version patch (e.g.: Ubuntu_20.04.*). It adds a filter to the patch version witch masks any patch version number.

### DIFF
--- a/agent/installer/installer.go
+++ b/agent/installer/installer.go
@@ -53,6 +53,7 @@ func getSupportedRegistry(ob algo.OutputBuilder) registry {
 		// BYOH Bundle Repository. Associate bundle with installer
 		linuxDistro := "Ubuntu_20.04.1_x86-64"
 		addBundleInstaller(linuxDistro, "v1.22.*", &algo.Ubuntu20_4K8s1_22{})
+
 		/*
 		 * PLACEHOLDER - ADD MORE K8S VERSIONS HERE
 		 */
@@ -62,6 +63,7 @@ func getSupportedRegistry(ob algo.OutputBuilder) registry {
 
 		// Match concrete os version to repository os version
 		reg.AddOsFilter("Ubuntu_20.04.*_x86-64", linuxDistro)
+
 		/*
 		 * PLACEHOLDER - POINT MORE DISTRO VERSIONS
 		 */
--- a/agent/installer/installer.go
+++ b/agent/installer/installer.go
@@ -62,9 +62,7 @@ func getSupportedRegistry(k8sVer string, ob algo.OutputBuilder) registry {
 		 */
 
 		// Match any patch version of the specified Major & Minor K8s version
-		reg.AddK8sFilter("v1.21.*")
 		reg.AddK8sFilter("v1.22.*")
-		reg.AddK8sFilter("v1.23.*")
 
 		// Match concrete os version to repository os version
 		reg.AddOsFilter("Ubuntu_20.04.*_x86-64", linuxDistro)
--- a/agent/installer/installer.go
+++ b/agent/installer/installer.go
@@ -37,9 +37,6 @@ type installer struct {
 // getSupportedRegistry returns a registry with installers for the supported OS and K8s
 func getSupportedRegistry(k8sVer string, ob algo.OutputBuilder) registry {
 	reg := newRegistry()
-	if k8sVer == "" {
-		k8sVer = "v1.22.3"
-	}
 
 	addBundleInstaller := func(osBundle, k8sBundle string, stepProvider algo.K8sStepProvider) {
 		a := algo.BaseK8sInstaller{
@@ -55,12 +52,16 @@ func getSupportedRegistry(k8sVer string, ob algo.OutputBuilder) registry {
 
 		// BYOH Bundle Repository. Associate bundle with installer
 		linuxDistro := "Ubuntu_20.04.1_x86-64"
+		// If no version specified use the minimum v1.22 supported
+		if k8sVer == "" {
+			k8sVer = "v1.22"
+		}
 		addBundleInstaller(linuxDistro, k8sVer, &algo.Ubuntu20_4K8s1_22{})
 		/*
 		 * PLACEHOLDER - ADD MORE K8S VERSIONS HERE
 		 */
 
-		// Match concrete k8s version to repository k8s version
+		// Match any patch version of the specified Major & Minor K8s version
 		reg.AddK8sFilter("v1.22.*")
 
 		// Match concrete os version to repository os version
--- a/agent/installer/installer.go
+++ b/agent/installer/installer.go
@@ -62,7 +62,9 @@ func getSupportedRegistry(k8sVer string, ob algo.OutputBuilder) registry {
 		 */
 
 		// Match any patch version of the specified Major & Minor K8s version
+		reg.AddK8sFilter("v1.21.*")
 		reg.AddK8sFilter("v1.22.*")
+		reg.AddK8sFilter("v1.23.*")
 
 		// Match concrete os version to repository os version
 		reg.AddOsFilter("Ubuntu_20.04.*_x86-64", linuxDistro)
--- a/agent/installer/installer.go
+++ b/agent/installer/installer.go
@@ -52,14 +52,10 @@ func getSupportedRegistry(ob algo.OutputBuilder) registry {
 
 		// BYOH Bundle Repository. Associate bundle with installer
 		linuxDistro := "Ubuntu_20.04.1_x86-64"
-		k8sVer := "v1.22.3"
-		addBundleInstaller(linuxDistro, k8sVer, &algo.Ubuntu20_4K8s1_22{})
+		addBundleInstaller(linuxDistro, "v1.22.3", &algo.Ubuntu20_4K8s1_22{})
 		/*
 		 * PLACEHOLDER - ADD MORE K8S VERSIONS HERE
 		 */
-
-		// Match concrete k8s version to repository k8s version
-		reg.AddK8sFilter("v1.22.*", k8sVer)
 
 		// Match concrete os version to repository os version
 		reg.AddOsFilter("Ubuntu_20.04.*_x86-64", linuxDistro)
--- a/agent/installer/installer.go
+++ b/agent/installer/installer.go
@@ -52,10 +52,14 @@ func getSupportedRegistry(ob algo.OutputBuilder) registry {
 
 		// BYOH Bundle Repository. Associate bundle with installer
 		linuxDistro := "Ubuntu_20.04.1_x86-64"
-		addBundleInstaller(linuxDistro, "v1.22.3", &algo.Ubuntu20_4K8s1_22{})
+		k8sVer := "v1.22.3"
+		addBundleInstaller(linuxDistro, k8sVer, &algo.Ubuntu20_4K8s1_22{})
 		/*
 		 * PLACEHOLDER - ADD MORE K8S VERSIONS HERE
 		 */
+
+		// Match concrete k8s version to repository k8s version
+		reg.AddK8sFilter("v1.22.*", k8sVer)
 
 		// Match concrete os version to repository os version
 		reg.AddOsFilter("Ubuntu_20.04.*_x86-64", linuxDistro)
--- a/agent/installer/installer.go
+++ b/agent/installer/installer.go
@@ -37,6 +37,9 @@ type installer struct {
 // getSupportedRegistry returns a registry with installers for the supported OS and K8s
 func getSupportedRegistry(k8sVer string, ob algo.OutputBuilder) registry {
 	reg := newRegistry()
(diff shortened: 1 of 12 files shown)


agent/installer/installer.go
```diff
@@ -52,11 +52,15 @@ func getSupportedRegistry(ob algo.OutputBuilder) registry {
 
 		// BYOH Bundle Repository. Associate bundle with installer
 		linuxDistro := "Ubuntu_20.04.1_x86-64"
-		addBundleInstaller(linuxDistro, "v1.22.3", &algo.Ubuntu20_4K8s1_22{})
+		k8sVer := "v1.22.3"
```
anusha94, line 55: Why is this hard coding still needed?

agent/installer/registry.go
```diff
@@ -11,18 +11,24 @@ import (
 type osk8sInstaller interface{}
 type k8sInstallerMap map[string]osk8sInstaller
 type osk8sInstallerMap map[string]k8sInstallerMap
-type filterBundlePair struct {
+type filterOsBundleMap struct {
 	osFilter string
 	osBundle string
 }
-type filterBundleList []filterBundlePair
+type filterK8sBundleMap struct {
```
anusha94, line 18: Suggestion - if we can use only `k8sFilter`, we probably would not need a dummy `v1.22.3` version to be mapped. This makes sense for OS, because we are treating all `20.04.*` as `20.04.1`, but we cannot map the same when it comes to k8s version.

I could be wrong though - I am trying to avoid the hardcoding of `v1.22.3` in `addBundleInstaller` call.

agent/installer/registry_test.go
```diff
@@ -41,6 +41,9 @@ var _ = Describe("Byohost Installer Tests", func() {
 			Expect(func() { r.AddBundleInstaller("rhel", "1.24", dummy124) }).NotTo(Panic())
 			r.AddOsFilter("ubuntu.*", "ubuntu")
 			r.AddOsFilter("rhel.*", "rhel")
+			r.AddK8sFilter("1.22.*", "1.22")
```
anusha94, line 44: Question - why are we not using complete semver of the k8s version? like `v1.22.3` or `v1.23.5`

agent/installer/registry.go
jeuxdemains: Ah, but of course. We don't need a map at all. A simple array with filter(s) will do just fine. Thanks, Anusha!

jeuxdemains:
I removed the hardcoded K8s version in the addBundleInstaller(); Also I refactored the K8s map to an array of filters. Updated the e2e tests as well.

I however have the feeling that my implementation is not good.
The addBundleInstaller() introduces a new argument which is the k8s version -- whenever it is empty it defaults to v1.22.3. I really don't like that and I will need your input for better implementation that will conform the usage of the API by the rest of the TKG components.

I'm very open for comments and opinions.

agent/installer/registry.go
jeuxdemains: You're totally right. I made the updates accordingly. However, it would be great if you have a look again because I don't feel that I came up with the best solution. I'm open for suggestions.

agent/installer/installer.go
```diff
@@ -52,11 +55,14 @@ func getSupportedRegistry(ob algo.OutputBuilder) registry {
 
 		// BYOH Bundle Repository. Associate bundle with installer
 		linuxDistro := "Ubuntu_20.04.1_x86-64"
-		addBundleInstaller(linuxDistro, "v1.22.3", &algo.Ubuntu20_4K8s1_22{})
+		addBundleInstaller(linuxDistro, k8sVer, &algo.Ubuntu20_4K8s1_22{})
```
georgievaVMW, line 58: I think that the K8s version needs to be specific value and not a variable in this case.
At the moment if well call the method with k8sVer = "1.24" it will again use the bundle installer for "1.22" although different steps might be needed.
As we discussed in our meeting, maybe currently the best option is to keep the version hardcoded but to change from "v1.22.3" to "v1.22"

jeuxdemains: This is something that I overlooked. Alex is right. As we are calling the algorithm for Ubuntu 20.04, Kubernetes 1.22 (algo.Ubuntu20_4K8s1_22), we are bound to hardcode both the OS & K8s major and minor versions. Maybe not directly as a string but it has to always match the algorithm version.
Thank you Alex for your remark.

anusha94: The algorithm if I understand correctly is just to indicate the sequence of steps. These need not be different for every OS-K8s combination. Is that correct?

Can we do something like -
`addBundleInstaller(linuxDistro, "v1.22.*", &algo.Ubuntu20_4K8s1_22{})`

So that in future it is easier to support other minor versions, like -
`addBundleInstaller(linuxDistro, "v1.23.*", &algo.Ubuntu20_4K8s1_22{})`
Note that the algo is still `1_22` which I think should be fine.

agent/installer/installer.go
```diff
@@ -35,8 +35,11 @@ type installer struct {
 }
 
 // getSupportedRegistry returns a registry with installers for the supported OS and K8s
-func getSupportedRegistry(ob algo.OutputBuilder) registry {
+func getSupportedRegistry(k8sVer string, ob algo.OutputBuilder) registry {
 	reg := newRegistry()
+	if k8sVer == "" {
+		k8sVer = "v1.22.3"
```
anusha94, line 41: This again is hardcoding to a particular patch version. Can we make this `v1.22.*`?

agent/installer/installer.go
jeuxdemains: Consider it done. Thanks both Anusha and Alex.

agent/installer/installer.go
```diff
@@ -52,11 +52,18 @@ func getSupportedRegistry(ob algo.OutputBuilder) registry {
 
 		// BYOH Bundle Repository. Associate bundle with installer
 		linuxDistro := "Ubuntu_20.04.1_x86-64"
-		addBundleInstaller(linuxDistro, "v1.22.3", &algo.Ubuntu20_4K8s1_22{})
+		// If no version specified use the minimum v1.22 supported
+		if k8sVer == "" {
```
anusha94, line 56: We don't have to do this. How about removing `k8sVer` argument from the function itself? Any consequences?

agent/installer/installer.go
```diff
@@ -52,11 +52,18 @@ func getSupportedRegistry(ob algo.OutputBuilder) registry {
...
+			k8sVer = "v1.22"
+		}
+		addBundleInstaller(linuxDistro, k8sVer, &algo.Ubuntu20_4K8s1_22{})
```
anusha94, line 59: Here, we can hard code to a particular minor series - 
`addBundleInstaller(linuxDistro, "v1.22.*", &algo.Ubuntu20_4K8s1_22{})`

agent/installer/installer.go
jeuxdemains: No consequences at all, actually I introduced this var in this update in order to apply variable matching. I will update it by reverting some changes now and will do the test. Thanks!

georgievaVMW (approved):
LGTM

agent/installer/registry.go
```diff
@@ -11,18 +11,20 @@ import (
 type osk8sInstaller interface{}
 type k8sInstallerMap map[string]osk8sInstaller
 type osk8sInstallerMap map[string]k8sInstallerMap
-type filterBundlePair struct {
+type filterOsBundleMap struct {
```
anusha94, line 14: Why'd you rename this as a map? it isn't a map really.

agent/installer/registry_test.go
```diff
@@ -36,21 +36,24 @@ var _ = Describe("Byohost Installer Tests", func() {
 			Expect(r.GetInstaller("a", "b")).To(BeNil())
 		})
 		It("Should allow working with installers", func() {
-			Expect(func() { r.AddBundleInstaller("ubuntu", "1.22", dummy122) }).NotTo(Panic())
-			Expect(func() { r.AddBundleInstaller("ubuntu", "1.23", dummy123) }).NotTo(Panic())
-			Expect(func() { r.AddBundleInstaller("rhel", "1.24", dummy124) }).NotTo(Panic())
+			Expect(func() { r.AddBundleInstaller("ubuntu", "v1.22.*", dummy122) }).NotTo(Panic())
+			Expect(func() { r.AddBundleInstaller("ubuntu", "v1.23.*", dummy123) }).NotTo(Panic())
```
anusha94, line 40: At this point we don't support `v1.23.*` and `v1.24.*`
These are negative test cases IMO. I understand the assertion is not to panic, but ideally anything other than `v1.22.*` should error out. Do we have such a test case?

agent/installer/registry.go
```diff
@@ -11,18 +11,20 @@ import (
...
+
+type filterBundleList []filterOsBundleMap
 
 // Registry contains
 // 1. Entries associating BYOH Bundle i.e. (OS,K8sVersion) in the Repository with Installer in Host Agent
 // 2. Entries that match a concrete OS to a BYOH Bundle OS from the Repository
 type registry struct {
 	osk8sInstallerMap
 	filterBundleList
+	k8sFilters []string
```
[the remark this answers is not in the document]
anusha94: Also update the comment for the registry struct now that it will include a new field

agent/installer/registry_test.go
jeuxdemains: We don't have test like that, I will convert those to such. Thanks.

agent/installer/registry.go
```diff
@@ -11,18 +11,26 @@ import (
...
 	osFilter string
 	osBundle string
 }
-type filterBundleList []filterBundlePair
+
+type filterK8sBundle struct {
+	k8sFilter string
+}
+
+type filterBundleList []filterOsBundlePair
```
anusha94, line 23:
```suggestion
type filterOSBundleList []filterOsBundlePair
```

Feel free to include this in one of the future PRs

anusha94 (approved):
lgtm.

agent/installer/registry.go
jeuxdemains: Will do it right away, thanks!

anusha94 (approved):
Thank you for this PR 🎉

agent/installer/registry.go
```diff
@@ -11,18 +11,26 @@ import (
...
-type filterBundlePair struct {
+type filterOsBundlePair struct {
 	osFilter string
 	osBundle string
 }
-type filterBundleList []filterBundlePair
+
+type filterK8sBundle struct {
+	k8sFilter string
+}
```
dharmjit, line 21: Question: Is there an advantage of defining a struct type to hold only a single string field?
could we instead have?
```
filterk8s string
```

dharmjit (commented):
lgtm, only a basic nit.

agent/installer/registry.go
jeuxdemains: It was like that at the beginning but with Anusha we decided to put it into a struct in order to be consistent with the OS bundle struct.

dharmjit (approved):

Issue #414, opened 2022-03-02
Update installer cli to indicate correct support matrix


anusha94:
**Describe the solution you'd like**
Current output of listing supported OS and K8s version
```
$ ./cli --list-supported
OS			K8S Version	BYOH Bundle Name
---			-----------	----------------
Ubuntu_20.04.1_x86-64	 v1.22.3	byoh-bundle-ubuntu_20.04.1_x86-64_k8s_v1.22.3
```

- We need to update the OS support to indicate `Ubuntu 20.04.*`
- After #411, we will be flexible enough to support any version in a given minor series. So the K8s version column should read `v1.22.*`. The output should contain a message saying the corresponding bundles (particular to a patch version) should be pushed to the OCI registry of choice. By default, BYOH uses [projects.registry.vmware.com](https://projects.registry.vmware.com/)
- Similarly, the BYOH Bundle Name also needs to be updated.
- Update [local-dev.md](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/main/docs/local_dev.md#cli) guide to reflect the changes.

**Anything else you would like to add:**
While we may have to wait for #411 to be completed to update K8s version column, in the meantime (as part of a different PR), update the OS column

dharmjit:
Closing this as it was fixed in this PR #437

PR #437, opened 2022-03-16
Handle bundle artifact naming changes

dharmjit:
This PR will handle the newer BYOH Bundles naming convention i.e. `byoh-bundle-ubuntu_20.04.1_x86-64_<bundleType>:<k8sVersion>` where

`bundleType` - k8s
`k8sVersion` - example - v1.22.3 

This PR introduces
- `bundleType` type to support future use cases of various bundle types 
- handles `--list-supported` output for installer cli
```
The corresponding bundles (particular to a patch version) should be pushed to the OCI registry of choice
By default, BYOH uses projects.registry.vmware.com

Note: It may happen that a specific patch version of a k8s minor release is not available in the OCI registry

OS                      K8S Version     BYOH Bundle Name
---                     -----------     ----------------
Ubuntu_20.04.*_x86-64    v1.21.*        byoh-bundle-ubuntu_20.04.1_x86-64_k8s:v1.21.*
Ubuntu_20.04.*_x86-64    v1.22.*        byoh-bundle-ubuntu_20.04.1_x86-64_k8s:v1.22.*
Ubuntu_20.04.*_x86-64    v1.23.*        byoh-bundle-ubuntu_20.04.1_x86-64_k8s:v1.23.*
```
**Which issue(s) this PR fixes**
fixes: #448 #414 

- This is a backward-incompatible change.
- Naming still uses the ubuntu patch version i.e. `20.04.1`  to map any ubuntu OS 20.04 patch version to the mentioned artifact.

### DIFF
--- a/agent/host_agent_suite_test.go
+++ b/agent/host_agent_suite_test.go
@@ -46,7 +46,7 @@ var (
 
 const (
 	bundleLookupBaseRegistry = "projects.registry.vmware.com/cluster_api_provider_bringyourownhost"
-	BundleLookupTag          = "v0.1.0_alpha.2"
+	BundleLookupTag          = "v1.22.3"
 	K8sVersion               = "v1.22.3"
 )
 
--- a/agent/installer/installer.go
+++ b/agent/installer/installer.go
@@ -32,6 +32,14 @@ const (
 	ErrBundleUninstall = Error("Error uninstalling bundle")
 )
 
+// BundleType is used to support various bundles
+type bundleType string
+
+const (
+	// BundleTypeK8s represents a vanilla k8s bundle
+	BundleTypeK8s bundleType = "k8s"
+)
+
 var preRequisitePackages = []string{"socat", "ebtables", "ethtool", "conntrack"}
 
 type installer struct {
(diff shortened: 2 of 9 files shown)


agent/host_agent_suite_test.go
```diff
@@ -46,7 +46,7 @@ var (
 
 const (
 	bundleLookupBaseRegistry = "projects.registry.vmware.com/cluster_api_provider_bringyourownhost"
-	BundleLookupTag          = "v0.1.0_alpha.2"
+	BundleLookupTag          = "v1.22.3"
```
anusha94, line 49: do we need this variable? We could reuse K8sVersion itself

agent/installer/installer.go
```diff
@@ -32,6 +32,14 @@ const (
 	ErrBundleUninstall = Error("Error uninstalling bundle")
 )
 
+// BundleType is used to support various bundles
+type bundleType string
+
+const (
+	// BundleTypeK8s represents a vanilla k8s bundle
+	BundleTypeK8s bundleType = "k8s"
```
anusha94, line 40: I'm assuming in future we would get this field from cluster-template and we remove this `const`, correct?

anusha94 (approved):
lgtm. minor nits.

agent/host_agent_suite_test.go
dharmjit: Spec still has both and didn't change it here.

agent/installer/installer.go
dharmjit: yup some refactoring will be required at that point in time, but this could be used in case there is no 1:1 mapping.

agent/host_agent_suite_test.go
anusha94: ah okay.

jeuxdemains:
lgtm. great stuff

PR #440, opened 2022-03-21
Added support for K8s minor versions v1.21.* & v1.23.* to the Installer 

### LINKED ISSUE #412: Support K8s v1.21, v1.23 minor series
This is a follow-up of https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/issues/411
Once we have the flexibility of supporting an entire minor series, we should expand our support to `v1.21` and `v1.23` series as well.

Whatever is the latest version of the minor series at the time of PR implementation, make sure to push the corresponding BYOH bundle to the [OCI registry](https://projects.registry.vmware.com/harbor/projects). Please contact the maintainers if you need help with pushing the bundles.

Kubernetes Release Dates and latest info can be found here - https://kubernetes.io/releases/

jeuxdemains:
This PR adds support to the Installer for Kubernetes minor versions v1.21.* and v1.23.*
Note that both of those minor versions have no differences in their steps provider so they are using the same installation algorithm as v1.22.*
  
Fixes #412

### DIFF
--- a/agent/installer/installer.go
+++ b/agent/installer/installer.go
@@ -52,14 +52,18 @@ func getSupportedRegistry(ob algo.OutputBuilder) registry {
 
 		// BYOH Bundle Repository. Associate bundle with installer
 		linuxDistro := "Ubuntu_20.04.1_x86-64"
+		addBundleInstaller(linuxDistro, "v1.21.*", &algo.Ubuntu20_4K8s1_22{})
 		addBundleInstaller(linuxDistro, "v1.22.*", &algo.Ubuntu20_4K8s1_22{})
+		addBundleInstaller(linuxDistro, "v1.23.*", &algo.Ubuntu20_4K8s1_22{})
 
 		/*
 		 * PLACEHOLDER - ADD MORE K8S VERSIONS HERE
 		 */
 
 		// Match any patch version of the specified Major & Minor K8s version
+		reg.AddK8sFilter("v1.21.*")
 		reg.AddK8sFilter("v1.22.*")
+		reg.AddK8sFilter("v1.23.*")
 
 		// Match concrete os version to repository os version
 		reg.AddOsFilter("Ubuntu_20.04.*_x86-64", linuxDistro)
--- a/agent/installer/installer.go
+++ b/agent/installer/installer.go
@@ -53,6 +53,7 @@ func getSupportedRegistry(ob algo.OutputBuilder) registry {
 		// BYOH Bundle Repository. Associate bundle with installer
 		linuxDistro := "Ubuntu_20.04.1_x86-64"
 		addBundleInstaller(linuxDistro, "v1.22.*", &algo.Ubuntu20_4K8s1_22{})
+
 		/*
 		 * PLACEHOLDER - ADD MORE K8S VERSIONS HERE
 		 */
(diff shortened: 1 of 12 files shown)


agent/installer/installer.go
```diff
@@ -52,14 +52,18 @@ func getSupportedRegistry(ob algo.OutputBuilder) registry {
 
 		// BYOH Bundle Repository. Associate bundle with installer
 		linuxDistro := "Ubuntu_20.04.1_x86-64"
+		addBundleInstaller(linuxDistro, "v1.21.*", &algo.Ubuntu20_4K8s1_22{})
```
dharmjit, line 55: Question: Do we need to use a specific minor k8s version in the name. Or we could use a type alias
```
type Ubuntu2004_k8s algo.Ubuntu20_4K8s1_22
```

jeuxdemains: We can do that, sounds good and reasonable. @anusha94, @georgievaVMW  what is your opinion?

anusha94: > Do we need to use a specific minor k8s version in the name

@dharmjit @jeuxdemains 
Good suggestion and I think this will work for now. The reason for including minor in the name was what if certain steps / sequence of steps change for a future K8s version? Then we would want to define another algo that has the new set of steps. But this could be rare, so we will end up using the same `Ubuntu20_4K8s1_22` algo as long as the steps to install k8s remain the same.
I'd leave it to you folks to decide, don't have a strong opinion :)

agent/installer/installer.go
```diff
@@ -52,14 +52,18 @@ func getSupportedRegistry(ob algo.OutputBuilder) registry {
...
+		addBundleInstaller(linuxDistro, "v1.23.*", &algo.Ubuntu20_4K8s1_22{})
 
 		/*
 		 * PLACEHOLDER - ADD MORE K8S VERSIONS HERE
 		 */
 
 		// Match any patch version of the specified Major & Minor K8s version
+		reg.AddK8sFilter("v1.21.*")
 		reg.AddK8sFilter("v1.22.*")
+		reg.AddK8sFilter("v1.23.*")
```
anusha94, line 66: unit tests please? :)

jeuxdemains: We have unit tests for the filters from my previous PR. If you have something more specific  in mind please let me know and I will code it.

dharmjit (approved):
lgtm

agent/installer/installer.go
anusha94: The current tests are for `v1.22.*` minor series. Ideally there should be 2 kinds of tests. One specifically for minor series support - something like if K8sFilter is set to`v1.foo.*`, it should not error on installing `v1.foo.bar`.

This PR specifically talks about adding support for a concrete minor version - `21.*` and `23.*` (without you having to add the filter yourself, the registry should accommodate these series)

Thoughts? The reason being there are no backing tests for the code / filters that you introduced. If I were to accidentally remove `reg.AddK8sFilter("v1.21.*")`, there would be no signal screaming that this is not allowed.

anusha94: This might involve structuring the current tests a bit differently. If you want to take this up in a follow-up PR, I'm good with it as we don't want to block @dharmjit :)

jeuxdemains: Okay, let's not block @dharmjit  yet however I will start working on a new PR that is implementing the tests right away. 
Thanks!

anusha94: Cool, please open a corresponding issue to track this effort.

anusha94 (approved):
Unit tests in a subsequent PR

PR #441, opened 2022-03-21
Wait for the host agent process to start in host agent integration tests

### LINKED ISSUE #438: Flaky Host Agent Integration tests
The host agent integration tests are quite flaky and are intermittently failing in the CI as well as local environments.
Here is one such failing CI example -  https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/runs/5570549277?check_suite_focus=true

The tests in this file need to revisited - https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/main/agent/host_agent_test.go

anusha94:
Since we are running the host agent in the containers for integration
tests, the process is not started instantly sometimes and this causes
flakes in the tests. Adding a check to ensure the host agent process is
started and only then proceed with the tests

Signed-off-by: Nilanjan Daw <dawn@vmware.com>

Fixes #438

### DIFF
--- a/agent/host_agent_test.go
+++ b/agent/host_agent_test.go
@@ -157,6 +157,18 @@ var _ = Describe("Agent", func() {
 
 			output, _, err = runner.ExecByoDockerHost(byoHostContainer)
 			Expect(err).NotTo(HaveOccurred())
+
+			// wait until the agent process starts inside the byoh host container
+			Eventually(func() bool {
+				containerTop, _ := runner.DockerClient.ContainerTop(ctx, byoHostContainer.ID, []string{})
+				for _, proc := range containerTop.Processes {
+					if strings.Contains(proc[len(containerTop.Titles)-1], "agent") {
+						return true
+					}
+
+				}
+				return false
+			}).Should(BeTrue())
 		})
 
 		AfterEach(func() {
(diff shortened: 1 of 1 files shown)


agent/host_agent_test.go
```diff
@@ -157,6 +157,18 @@ var _ = Describe("Agent", func() {
...
+			Eventually(func() bool {
+				containerTop, _ := runner.DockerClient.ContainerTop(ctx, byoHostContainer.ID, []string{})
+				for _, proc := range containerTop.Processes {
+					if strings.Contains(proc[len(containerTop.Titles)-1], "agent") {
+						return true
+					}
+
+				}
+				return false
+			}).Should(BeTrue())
```
dharmjit, line 171: Do we set an explicit timeout interval for this `eventually` block? This might fail as the default timeout is 1 sec.

anusha94: Good point. What do you think is a reasonable timeout for this? `60s`?

dharmjit (approved):
lgtm

jamiemonserrate:
I think this makes sense, and is a good incremental step to improving the flakiness. I am _slightly_ worried that this check might pass wrongly if there is _any_ process that matches the name `agent`. But this is a good start. 

But is this enough? That is, should we be checking if the host-agent has reached a desired state? Like maybe the log messages should say "agent started/ready" or something?

Just a suggestion, as I'm not sure what exactly we need to wait for.

jamiemonserrate (approved):

anusha94:
> But is this enough? That is, should we be checking if the host-agent has reached a desired state? Like maybe the log messages should say "agent started/ready" or something?

Good suggestion, if the flakiness persists, we could try that.

PR #449, opened 2022-03-22
Fix for bundle version in the installer CLI

georgievaVMW:
Made small changes so the CLI shows correct bundle version.
Made a map from k8s filter to k8s bundle (this is in order to easily output correct bundle for example 1.23.* -> 1.22)
Made changes to installer.go and the registry tests to accommodate the added second argument to the function.

### DIFF
--- a/agent/installer/cli-dev.go
+++ b/agent/installer/cli-dev.go
@@ -107,7 +107,7 @@ func listSupported() {
 	osFilters, osBundles := ListSupportedOS()
 	for i := range osFilters {
 		for _, k8s := range ListSupportedK8s(osBundles[i]) {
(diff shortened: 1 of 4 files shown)


jeuxdemains:
A note on why we switched from filter array to filter map.
We had a discussion with Alex about this update and the path he can take to implement it.

Initially the installer was meant to use a simple container (array) to hold the filters and that was the right choice at the time.
However, the CLI require certain level of predictability in order to perform as expected (to print the supported versions) which is not possible without predefining to which K8s versions the filters match.

We discussed different solutions like stripping the filter masks (the dot asterisk part) from the string of the help of the CLI, as well as extracting the map to separate module which could be shared between modules (CLI and Installer). 
Ultimately we concluded integrating it to the Installer as a map would cost less time and be better prepared for the future, requiring less portions of the code to be maintained.

PR #451, opened 2022-03-22
Installer: e2e tests against unsupported minor & major versions of K8s

### LINKED ISSUE #450: Installer K8s filters integration tests
With the latest update we introduced K8s filters for patch versions (e.g.: 1.22.x).
We added support for v1.21.x and v1.23.x, now we need to implement integration tests for them.

jeuxdemains:

Implements e2e test against unsupported minor and major K8s versions of the Installer.
The test should fail on any unsupported versions.

Fixes #450


agent/installer/installer_test.go
```diff
@@ -40,6 +40,31 @@ var _ = Describe("Byohost Installer Tests", func() {
 			}
 		})
 	})
+	Context("When installer is created", func() {
```
anusha94, line 43: This and the below Contexts read the same. You can choose to either set different description to Contexts (so that they are meaningful for the tests they contain), or if they are really the same, then move the `It` blocks under a single Context

anusha94 (approved):
Lgtm

agent/installer/installer_test.go
jeuxdemains: done. B)

dharmjit (approved):
lgtm

Issue #415, opened 2022-03-02
Documentation for pushing BYOH bundles to OCI Registry


anusha94:
**Describe the solution you'd like**
For the Host Agent Installer to work, we need BYOH K8s bundles to be present in an OCI Registry. Currently, we have a few scripts to generate these bundles. But there is no instructions provided for pushing these bundles.
It will be good to have a doc outlining how to generate and push BYOH Bundles to OCI registries.

Bundle builder scripts can be found here - https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/tree/main/agent/installer/bundle_builder

PR #455, opened 2022-03-24
Update k8s version in e2e tests

### LINKED ISSUE #413: Update e2e test to use latest K8s version
Currently, our e2e tests are pinned to K8s `v1.22.3` version. That's because we only have `v1.22.3` support in the installer. With #411 , we intend to make the installer more flexible. We should therefore move to testing higher K8s version.

**Explore**
Which K8s version needs to be tested and how frequently we have to update the tests can be tricky. For this, let's refer to CAPI and other providers' e2e tests to see what the current pattern of testing is. We can then make a call on which K8s version to pin to.

This issue cannot be worked upon until #411 is completed. But in the meantime, it is good to do the exploration of identifying the K8s version in e2e tests.

anusha94:
- updated e2e k8s version to `v1.23.5`
- updated corresponding Dockerfiles
- pushed a new bundle to Harbor
- minor corrections to docs

Fixes #413 #415 

**Additional information**
New OCI image pushed to Harbor. You can check with `docker pull projects.registry.vmware.com/cluster_api_provider_bringyourownhost/byoh-bundle-ubuntu_20.04.1_x86-64_k8s:v1.23.5`

### DIFF
--- a/agent/installer/bundle_builder/ingredients/deb/Dockerfile
+++ b/agent/installer/bundle_builder/ingredients/deb/Dockerfile
@@ -8,12 +8,12 @@
 # 2. Run the image
 #
 
-ARG BASE_IMAGE=ubuntu:18.04
+ARG BASE_IMAGE=ubuntu:20.04
 FROM $BASE_IMAGE as build
 
 # Override to download other version
-ENV CONTAINERD_VERSION=1.5.7
-ENV KUBERNETES_VERSION=1.21.2-00
+ENV CONTAINERD_VERSION=1.6.0
+ENV KUBERNETES_VERSION=1.23.5-00
 ENV ARCH=amd64
 
 RUN apt-get update \
--- a/Makefile
+++ b/Makefile
@@ -10,8 +10,8 @@ RELEASE_DIR := _dist
 
 # Image URL to use all building/pushing image targets
 IMG ?= ${STAGING_REGISTRY}/${IMAGE_NAME}:${TAG}
-BYOH_BASE_IMG = byoh/node:v1.22.3
-BYOH_BASE_IMG_DEV = byoh-dev/node:v1.22.3
+BYOH_BASE_IMG = byoh/node:e2e
+BYOH_BASE_IMG_DEV = byoh/node:dev
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
 CRD_OPTIONS ?= "crd:trivialVersions=true,preserveUnknownFields=false"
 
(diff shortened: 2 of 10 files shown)


agent/installer/bundle_builder/ingredients/deb/Dockerfile
```diff
@@ -12,8 +12,8 @@ ARG BASE_IMAGE=ubuntu:18.04
 FROM $BASE_IMAGE as build
 
 # Override to download other version
-ENV CONTAINERD_VERSION=1.5.7
-ENV KUBERNETES_VERSION=1.21.2-00
+ENV CONTAINERD_VERSION=1.6.0
+ENV KUBERNETES_VERSION=1.23.4-00
```
shamsher31, line 16: Why there is a need to use `00` at the end of the version? I remember this is required for apt-get to fetch the correct version. but is it required for Docker to pull the image?

agent/installer/bundle_builder/ingredients/deb/Dockerfile
```diff
@@ -12,8 +12,8 @@ ARG BASE_IMAGE=ubuntu:18.04
 FROM $BASE_IMAGE as build
```
shamsher31, line 12: The `$BASE_IMAGE` still uses `ubuntu:18.04`. Is it ok to use `ubuntu:20.04` as this is the supported host os for BYOH?

anusha94: Done. Thanks

agent/installer/bundle_builder/ingredients/deb/Dockerfile
anusha94: These are not for pulling images. If you see the [download.sh](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/main/agent/installer/bundle_builder/ingredients/deb/download.sh) file, these variables are used to download the deb packages via `apt-get`

dharmjit (approved):
lgtm

dharmjit:
@anusha94, as `BYOH` is supporting `n-2` k8s minor versions, we need to push maybe the latest patch version of these minor k8s releases and also had to include e2e tests for different k8s versions. wdyt?

shubham14bajpai (approved):
lgtm

anusha94:
> we need to push maybe the latest patch version of these minor k8s releases and also had to include e2e tests for different k8s versions. wdyt?

@dharmjit I'm looking at jobs for CAPI - https://github.com/kubernetes/test-infra/tree/master/config/jobs/kubernetes-sigs/cluster-api
They seem to test the latest stable version in e2e and an upgrade test would test n-1 to n
So this seems sufficient to me now.

PR #477, opened 2022-04-05
Auto-detect the default BYOH interface for hosts

VibhorChinda:
**What this PR does / why we need it**:
Makes changes in the codebase to implement the auto-detect the default BYOH interface for hosts.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes # https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/issues/466

**Additional information**
Changed the template files [deleted the vim_interface env variable]

**Special notes for your reviewer**

Issue #288, opened 2021-12-21
Remove host from inventory


kerenbenzion:
I am trying to use byoh provider and I ran into an issue which I cannot delete an host which is not in use. 
I am trying to run uninstall of the agent on the host side , and to run delete on the management side and I get the following error: 

the command I am trying to run :
- kubectl delete byohosts.infrastructure.cluster.x-k8s.io lab-preprod-k8s-worker-lab-5

the result: 
- Error from server (Forbidden): admission webhook "vbyohost.kb.io" denied the request: byohost.infrastructure.cluster.x-k8s.io "lab-preprod-k8s-worker-lab-5" is forbidden: cannot delete ByoHost when MachineRef is assigned

I have also tried to edit the byoh host object , but I dont manage to commit any changes when I am deleting the machine ref. 

Can you let me know what should I do? 

- Kubernetes version: (use `kubectl version --short`): Client Version: v1.22.4
Server Version: v1.22.4
- OS: ubuntu 20.4

dharmjit:
Hi @kerenbenzion, There is a delete webhook that restricts deleting a `ByoHost` CR with a `status.MachineRef` pointing to a `ByoMachine`. The `Host-Agent` should be running and then you can try to scale out `MD`/`KCP`.

Note that removing a specific `ByoHost`/`ByoMachine` from the workload cluster is not supported as of now.

kerenbenzion:
@dharmjit  if I understood correctly, currently we can do only scale up ? We cannot decrease number of hosts if machine ref has been assigned already ? 
Will it be supported?

dharmjit:
>currently we can do only scale up

We can do both scale up/scale down for both `KCP`/`MD` as per the `CAPI` contract. you can use below to update the number of replicas
- `kubectl scale kcp|md`
- `kubectl edit kcp|md`

> We cannot decrease number of hosts.

If I understand your question correctly, you can detach hosts from the cluster with scale-out operation and once you have identified the host which is removed, you can stop the `Host-Agent` and delete the specific `ByoHost` CR.

kerenbenzion:
Hi, 

I am trying to understand if lets say one of my hosts (worker/controlplane) have issues (physical/os ) and I need some time to fix it 
Can I remove the byoh host untill the problem is fixed?

dharmjit:
> Can I remove the byoh host untill the problem is fixed?

Removing a specific host is not supported as of now.

[...]

PR #479, opened 2022-04-08
Platform specific fields in BYOHost CRD

### LINKED ISSUE #469: Add Host platform related fields to ByoHost CRD
Improve `ByoHost`CRD to include host platform details such as Operating System, Architecture, etc. In this issue, we could also discuss what all fields could be added and where it makes sense to keep this information(`Spec`, `Status`).

sachinkumarsingh092:
We need platform-specific info regarding the host in BYOHost CRDs. 

Fixes #469 

**Additional information**

Changes in brief: 
- **/apis/infrastructure/v1beta1/byohost_types.go**: We introduce new type `HostStatus` which has all the platformspecific details like OS name, Kernel version, architecture etc. We introduce a new field in `ByoHostStatus` with this type.
- **/config/crd/bases/infrastructure.cluster.x-k8s.io_byohosts.yaml**: Using `make generate`, we let controller-gen introduce these new fields in the CRDs automatically.
- **/apis/infrastructure/v1beta1/zz_generated.deepcopy.go**: Autogenerated implementation of the deepcopy methods for type `HostStatus`.

```
<none>
```


apis/infrastructure/v1beta1/byohost_types.go
```diff
@@ -32,6 +32,26 @@ type ByoHostSpec struct {
 	BootstrapSecret *corev1.ObjectReference `json:"bootstrapSecret,omitempty"`
 }
 
+// HostStatus is a set of details about the host platform.
+type HostStatus struct {
+	// MachineID reported by the host. For unique machine identification
+	// in the cluster this field is preferred. Learn more from man(5)
+	// machine-id: http://man7.org/linux/man-pages/man5/machine-id.5.html
+	HostID string `json:"hostid,omitempty"`
```
anusha94, line 40: Why do we need HostID and KernelVersion at this point?

sachinkumarsingh092: We discussed these fields in the corresponding issue (#479). 
We need some unique identifier like a machine-id (for Linux only) to identify the particular host in the workload pool. Similarly, if we already have an OS in place, having info about its kernel is useful in many cases when we need to know the compatibility of certain dependencies.
Makes sense?

anusha94: > We need some unique identifier like a machine-id (for Linux only) to identify the particular host in the workload pool

How is this information consumed by the provider or the user?

> having info about its kernel is useful in many cases when we need to know the compatibility of certain dependencies

What dependencies? I'd hold off adding this field until we have some need / requirement from the user.

apis/infrastructure/v1beta1/byohost_types.go
```diff
@@ -32,6 +32,26 @@ type ByoHostSpec struct {
...
+
+	// The Operating System reported by the host.
+	OSName string `json:"osname,omitempty"`
+
+	// OS Image reported by the host.
+	OSImage string `json:"osimage,omitempty"`
```
anusha94, line 46: Can you give an example value for this field?

apis/infrastructure/v1beta1/byohost_types.go
```diff
@@ -32,6 +32,26 @@ type ByoHostSpec struct {
 	BootstrapSecret *corev1.ObjectReference `json:"bootstrapSecret,omitempty"`
 }
 
+// HostStatus is a set of details about the host platform.
+type HostStatus struct {
```
anusha94, line 36: Where is the OS version stored? Will it be part of the OS name itself?

apis/infrastructure/v1beta1/byohost_types.go
sachinkumarsingh092: > How is this information consumed by the provider or the user?

Hmm. I was thinking in reference to targeting a specific host for something (like removing a specific host [#288]) but now I think we can just use the name of the particular host (as assigned after registration) for such things so we can remove this.

> What dependencies?

Like some driver support or some new improvements that are made in the latest versions that were not available in the previous kernel versions.

apis/infrastructure/v1beta1/byohost_types.go
sachinkumarsingh092: > Where is the OS version stored? Will it be part of the OS name itself?

Yes. 

I used [this ](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/issues/469#issuecomment-1091154744)example from the issue as the reference though some of the fields were removed or renamed according to our needs:

```
  nodeInfo:
    architecture: arm64
    bootID: 140d092c-9a92-4c56-b841-e1392d499a59
    containerRuntimeVersion: containerd://1.5.10
    kernelVersion: 5.4.0-105-generic
    kubeProxyVersion: v1.23.4
    kubeletVersion: v1.23.4
    machineID: 83fb87a59e494f40b7d93ed2ff8f23e1
    operatingSystem: linux
    osImage: Ubuntu 21.10
    systemUUID: 83fb87a59e494f40b7d93ed2ff8f23e1
```

apis/infrastructure/v1beta1/byohost_types.go
sachinkumarsingh092: OSName: linux
OSImage: Ubuntu 21.10

apis/infrastructure/v1beta1/byohost_types.go
dharmjit: Yup, I agree with @anusha94 and think we can introduce these 2 fields at a later point in time as and when required and in the PR let's just do the minimal
( I know adding these fields will provide more info about the host, but we are not sure how are we going to use this).

mayur-tolexo:
Just curious, why host `OS` and `Arch` details are not a part of `Spec` and added in `Status` of `ByoHost`?

sachinkumarsingh092:
> why host OS and Arch details are not a part of Spec and added in Status of ByoHost

@mayur-tolexo according to [kubebuilder docs](https://book.kubebuilder.io/cronjob-tutorial/api-design.html), `Spec` holds desired state, so any inputs to our controller go here while `Status` holds observed state and contains any information we want users or other controllers to be able to easily obtain.
By these definitions, OS and Arch details should be in `Status`.

mayur-tolexo:
@sachinkumarsingh092, thanks for clarifying. Ya make sense to provide only status `subresource` access to other controllers for `HostInfo`.

apis/infrastructure/v1beta1/byohost_types.go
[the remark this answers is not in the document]
sachinkumarsingh092: Oh! Just realized that `HostInfo` is also a struct used during registration ([host_register.go](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/main/agent/registration/host_registrar.go#L26)). But this will be removed by #477.

anusha94 (approved):
lgtm

dharmjit (approved):
lgtm

apis/infrastructure/v1beta1/byohost_types.go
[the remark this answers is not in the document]
dharmjit: Oh ok, ^^ type will be introduced in [byohost_types](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/main/apis/infrastructure/v1beta1/byohost_types.go) so it should be okay.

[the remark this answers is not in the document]
dharmjit: Thanks for highlighting this. I guess it should be okay.

PR #491, opened 2022-04-20
Sequence diagram from Bootstrap Token Authentication for BYO Host

### LINKED ISSUE #484: [Feature] Limit BYO Host agent from using admin credentials
Our current solution involves providing the host agent on every host with admin credentials, so that they can CRUD (Create Read Update Delete) a BYO Host CRD. This is less than ideal since every host now has access to all the Kubernetes API resources on the management cluster. 

I would like to have a solution which involves not sharing the admin credentials and instead something like a short-lived bootstrap token for the host authentication / registration process

anusha94:
This PR adds the sequence diagram for the Bootstrap Token Authentication for BYO Host workflow. This will allow the BYO Host to register into the capacity pool without having to use the admin creds of the management cluster.

Part of #484 

**Additional information**

Will convert this png to a puml diagram in a different PR


dharmjit (approved):
lgtm

pshail:
LGTM

we should add the uml file too for the same

pshail (approved):

sachinkumarsingh092 (approved):
lgtm

PR #495, opened 2022-04-20
installation and uninstallation command added from agent installer algo

### LINKED ISSUE #486: Create installation / uninstallation scripts from agent installer algo
Create `installation` / `uninstallation` scripts from agent installer [algo](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/main/agent/installer/internal/algo/ubuntu20_4K8s1_22.go).    
We will be storing this algo as bash script and the new installer controller in #467 will be using these scripts to create the installation secret.

mayur-tolexo:

Added the consolidated steps that need to perform for installing/uninstalling of supported k8s version on supported os as variables in installer package

Fixes #486

### DIFF
--- a/agent/installer/steps.go
+++ b/agent/installer/steps.go
@@ -1,0 +1,63 @@
+// Copyright 2021 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package installer
+
+// contains the installation and uninstallation steps for the supported os and k8s
+var (
+	DoUbuntu20_4K8s1_22 = `
+set -euo pipefail
+
+BUNDLE_PATH=${BUNDLE_PATH:-"/var/lib/byoh/bundles"}
+
+## disable swap
+swapoff -a && sed -ri '/\sswap\s/s/^#?/#/' /etc/fstab
+
+## disable firewall
+ufw disable
+
+## load kernal modules
+modprobe overlay && modprobe br_netfilter
+
+## adding os configuration
+tar -C / -xvf "$BUNDLE_PATH/conf.tar" && sysctl --system 
+
+## installing deb packages
+for pkg in cri-tools kubernetes-cni kubectl kubeadm kubelet; do
+  dpkg --install "$BUNDLE_PATH/$pkg.deb" && apt-mark hold $pkg
+done
+
+## intalling containerd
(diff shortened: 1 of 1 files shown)


agent/installer/steps.go
```diff
@@ -0,0 +1,63 @@
...
+
+// contains the installation and uninstallation steps for the supported os and k8s
+var (
+	DoUbuntu20_4K8s1_22 = `
+set -euo pipefail
+
+BUNDLE_PATH=${BUNDLE_PATH:-"/var/lib/byoh/bundles"}
+
+## disable swap
+swapoff -a && sed -ri '/\sswap\s/s/^#?/#/' /etc/fstab
```
shamsher31, line 14: I was assuming these scripts will be generated by the agent installer and then assigned to `DoUbuntu20_4K8s1_22 ` variable.

Aren't we duplicating these scripts, one in the agent installer and the other in these variables?

mayur-tolexo: Agreed, for some time these script will be duplicated.

we are planing to update the byo agent to install the bundle two ways:
1. agent will install the way its currently doing
2. new `K8sInstallationConfig` controller will generate the installation script and agent will execute it

These variables will be used by new `K8sInstallationConfig` controller to create the secret.

agent/installer/steps.go
```diff
@@ -0,0 +1,63 @@
+// Copyright 2021 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package installer
+
+// contains the installation and uninstallation steps for the supported os and k8s
+var (
+	DoUbuntu20_4K8s1_22 = `
+set -euo pipefail
```
shamsher31, line 9: Do we also need to set `#!/usr/bin/env bash` at the top of the script as it will run as a bash script?

mayur-tolexo: we will be executing this script using `cmd := exec.Command("/bin/bash", DoUbuntu20_4K8s1_22)` so `#!/usr/bin/env bash` won't be  required here.

shamsher31 (approved):
lgtm

anusha94 (approved):
lgtm

anusha94:
We can also consider a different location for this file. Eventually, we want to move away from having much code in the byoh binary.
So maybe in a future PR.

PR #501, opened 2022-04-25
Add host platform to ByoHost CR info on registration

### LINKED ISSUE #470: Add ByoHost CR with the Host platform information
After #469, we need to add the host platform details while registering `ByoHost` CR in the Host-Agent.

shubham14bajpai:
Signed-off-by: Shubham Bajpai <shbajpai@vmware.com>

This PR adds the host platform information (arch/os/distribution) to the ByoHost CR when the agent registers the host. Tested the same locally and this is what the CR would look like.

```yaml
$ kubectl get byohosts host1 -oyaml
apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
kind: ByoHost
metadata:
  creationTimestamp: "2022-04-25T05:52:54Z"
  generation: 1
  name: host1
  namespace: default
  resourceVersion: "169327"
  uid: 572acde9-2d80-446c-ab7a-260f71e8e8f4
spec: {}
status:
  conditions:
  - lastTransitionTime: "2022-04-25T05:52:55Z"
    reason: WaitingForMachineRefToBeAssigned
    severity: Info
    status: "False"
    type: K8sNodeBootstrapSucceeded
  hostinfo:
    architecture: arm64
    osimage: Ubuntu 20.04.4 LTS
    osname: linux
  network:
  ...
```

This PR also adds additional printers for the Byohost CRD and with those the output would look like this:
```shell
$ kubectl get byohost                                                  
NAME    OSNAME   OSIMAGE              ARCH
host1   linux    Ubuntu 20.04.4 LTS   amd64
host2   linux    Ubuntu 20.04.4 LTS   amd64
```

Fixes #470 

**Additional information**

### DIFF
--- a/agent/registration/host_registrar.go
+++ b/agent/registration/host_registrar.go
@@ -5,9 +5,16 @@ package registration
 
 import (
 	"context"
+	"fmt"
+	"io/ioutil"
 	"net"
+	"os"
+	"regexp"
+	"runtime"
+	"strings"
 
 	"github.com/jackpal/gateway"
+	"github.com/pkg/errors"
 	infrastructurev1beta1 "github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/apis/infrastructure/v1beta1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -80,6 +87,11 @@ func (hr *HostRegistrar) UpdateNetwork(ctx context.Context, byoHost *infrastruct
 
 	byoHost.Status.Network = hr.GetNetworkStatus()
 
+	klog.Info("Add Host Info")
+	if byoHost.Status.HostDetails, err = hr.getHostInfo(); err != nil {
+		return err
+	}
+
 	return helper.Patch(ctx, byoHost)
 }
 
@@ -128,4 +140,38 @@ func (hr *HostRegistrar) GetNetworkStatus() []infrastructurev1beta1.NetworkStatu
 		Network = append(Network, netStatus)
 	}
 	return Network
+}
+
+// getHostInfo gets the host platform details...
+func (hr *HostRegistrar) getHostInfo() (infrastructurev1beta1.HostInfo, error) {
+	hostInfo := infrastructurev1beta1.HostInfo{}
+
+	hostInfo.Architecture = runtime.GOARCH
+	hostInfo.OSName = runtime.GOOS
+
+	if distribution, err := getOperatingSystem(); err != nil {
+		return hostInfo, errors.Wrap(err, "failed to get host operating system image")
+	} else {
+		hostInfo.OSImage = distribution
+	}
+	return hostInfo, nil
+}
+
+// getOperatingSystem gets the name of the current operating system image.
+func getOperatingSystem() (string, error) {
+	rex := regexp.MustCompile("(PRETTY_NAME)=(.*)")
+
+	bytes, err := ioutil.ReadFile("/etc/os-release")
+	if err != nil && os.IsNotExist(err) {
+		// /usr/lib/os-release in stateless systems like Clear Linux
+		bytes, err = ioutil.ReadFile("/usr/lib/os-release")
+	}
+	if err != nil {
+		return "", fmt.Errorf("error opening file : %v", err)
+	}
+	line := rex.FindAllStringSubmatch(string(bytes), -1)
+	if len(line) > 0 {
+		return strings.Trim(line[0][2], "\""), nil
+	}
+	return "Linux", nil
 }
--- a/apis/infrastructure/v1beta1/byohost_types.go
+++ b/apis/infrastructure/v1beta1/byohost_types.go
@@ -68,6 +68,9 @@ type ByoHostStatus struct {
 //+kubebuilder:object:root=true
 //+kubebuilder:resource:path=byohosts,scope=Namespaced,shortName=byoh
 //+kubebuilder:subresource:status
+//+kubebuilder:printcolumn:name="OSName",type="string",JSONPath=`.status.hostinfo.osname`
+//+kubebuilder:printcolumn:name="OSImage",type="string",JSONPath=`.status.hostinfo.osimage`
+//+kubebuilder:printcolumn:name="Arch",type="string",JSONPath=`.status.hostinfo.architecture`
 
 // ByoHost is the Schema for the byohosts API
 type ByoHost struct {
--- a/config/crd/bases/infrastructure.cluster.x-k8s.io_byohosts.yaml
+++ b/config/crd/bases/infrastructure.cluster.x-k8s.io_byohosts.yaml
@@ -18,7 +18,17 @@ spec:
     singular: byohost
   scope: Namespaced
   versions:
-  - name: v1beta1
+  - additionalPrinterColumns:
+    - jsonPath: .status.hostinfo.osname
+      name: OSName
+      type: string
+    - jsonPath: .status.hostinfo.osimage
+      name: OSImage
+      type: string
+    - jsonPath: .status.hostinfo.architecture
+      name: Arch
+      type: string
+    name: v1beta1
     schema:
       openAPIV3Schema:
         description: ByoHost is the Schema for the byohosts API


agent/registration/host_registrar.go
```diff
@@ -129,3 +141,37 @@ func (hr *HostRegistrar) GetNetworkStatus() []infrastructurev1beta1.NetworkStatu
...
+		bytes, err = ioutil.ReadFile("/usr/lib/os-release")
+	}
+	if err != nil {
+		return "", fmt.Errorf("error opening file : %v", err)
+	}
+	line := rex.FindAllStringSubmatch(string(bytes), -1)
+	if len(line) > 0 {
+		return strings.Trim(line[0][2], "\""), nil
+	}
+	return "Linux", nil
```
anusha94, line 176: Why are we defaulting here?

anusha94, line 176: Why are we defaulting here?

anusha94:
Also, add few tests :)

anusha94:
Also, add few tests :)

agent/registration/host_registrar.go
```diff
@@ -129,3 +141,37 @@ func (hr *HostRegistrar) GetNetworkStatus() []infrastructurev1beta1.NetworkStatu
 	}
 	return Network
 }
+
+// getHostInfo gets the host platform details...
```
sachinkumarsingh092, line 145:
```suggestion
// getHostInfo gets the host platform details.
```

agent/registration/host_registrar.go
```diff
@@ -80,6 +87,11 @@ func (hr *HostRegistrar) UpdateNetwork(ctx context.Context, byoHost *infrastruct
 
 	byoHost.Status.Network = hr.GetNetworkStatus()
 
+	klog.Info("Add Host Info")
```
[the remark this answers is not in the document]
shubham14bajpai: Makes sense as the information may change on os upgrades. Also we plan on adding cpu and other platform information.

agent/registration/host_registrar.go
shubham14bajpai: If for some reason the OS distribution does not support same file for the information marking it as generic linux. Picked it from kubelet code. We can also mark it as `Unknown`.

agent/registration/host_registrar.go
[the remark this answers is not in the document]
shubham14bajpai: But this will mean we need to call getHostInfo() on each reconcile. Is it okay to do so? As this info won't change very frequently.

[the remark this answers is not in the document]
shubham14bajpai: Yes that can be done it won't track OS upgrades like Ubuntu 20.04 to Ubuntu 21.10 etc.

agent/registration/host_registrar.go
anusha94: Alright. Thanks.

agent/reconciler/host_reconciler.go
```diff
@@ -311,6 +322,40 @@ func (r *HostReconciler) deleteEndpointIP(ctx context.Context, byoHost *infrastr
...
+	if distribution, err := getOperatingSystem(); err != nil {
+		return hostInfo, errors.Wrap(err, "failed to get host operating system image")
+	} else {
+		hostInfo.OSImage = distribution
+	}
+	return hostInfo, nil
+}
+
+// getOperatingSystem gets the name of the current operating system image.
+func getOperatingSystem() (string, error) {
```
anusha94, line 341: Are we re-inventing the wheel here? We have something similar that installer does in order to determine if pre-requisites are met. Check this -
https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/6ba32a1927018c06008a5cd38606cc35337a0065/agent/installer/os_detector.go#L26-L28

shubham14bajpai: Its somewhat similar but this returns a normalised string of os_version_arch, but we want these separately as different fields. We can either introduce similar fields to this struct and then reuse it or would have to re-parse the string again to break it into the fields. First option seems better. WDYT?

DavidSpek:
I don’t mean to hijack this PR, but thought I’d mention https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/proposals/20210310-opt-in-autoscaling-from-zero.md since it seems related and could possibly be included in this PR as well.

agent/reconciler/host_reconciler.go
shubham14bajpai: Made use of the existing os detector by adding a new function for this use case.

agent/reconciler/host_reconciler.go
```diff
@@ -92,6 +94,12 @@ func (r *HostReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctr
 
 func (r *HostReconciler) reconcileNormal(ctx context.Context, byoHost *infrastructurev1beta1.ByoHost) (ctrl.Result, error) {
 	logger := ctrl.LoggerFrom(ctx)
+	if err := r.populateHostDetails(ctx, byoHost); err != nil {
+		logger.Error(err, "error getting host platform details")
+		r.Recorder.Eventf(byoHost, corev1.EventTypeWarning, "GetHostInfoFailed", err.Error())
+		return ctrl.Result{}, err
+	}
+
```
dharmjit, line 102: Adding this inside reconcile loop doesn't feel right. Are there any strong opinions?

shubham14bajpai: This executes only once as the function checks for empty struct and then calls the os detector.

agent/installer/os_detector_test.go
```diff
@@ -75,35 +75,41 @@ var _ = Describe("Byohost Installer Tests", func() {
 			_, err = d.Detect()
 			Expect(err).ShouldNot(HaveOccurred())
 		})
+
+		It("Should return os name on GetOS", func() {
+			detectedOS, err = d.GetOSNameWithVersion(func() (string, error) { return mh.Get(os, ver, arch) })
+			Expect(err).ShouldNot(HaveOccurred())
+			Expect(detectedOS).To(Equal("Ubuntu 20.04.3"))
```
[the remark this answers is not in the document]
shubham14bajpai: Here we mock the hostnamectl output using variables and a mock function so this test will run any platform as we set those variable to these values https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/main/agent/installer/os_detector_test.go#L35-L38. But a better way would be check for `Equal(os + " " + ver)`

agent/reconciler/reconciler_test.go
```diff
@@ -94,6 +94,25 @@ var _ = Describe("Byohost Agent Tests", func() {
...
+			Expect(result).To(Equal(controllerruntime.Result{}))
+			Expect(reconcilerErr).ToNot(HaveOccurred())
+
+			updatedByoHost := &infrastructurev1beta1.ByoHost{}
+			err := k8sClient.Get(ctx, byoHostLookupKey, updatedByoHost)
+			Expect(err).ToNot(HaveOccurred())
+			hostDetails := updatedByoHost.Status.HostDetails
+			Expect(hostDetails).To(Equal(infrastructurev1beta1.HostInfo{
+				OSName:       "linux",
+				OSImage:      "Ubuntu 20.04.4",
```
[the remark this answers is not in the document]
shubham14bajpai: This version is taken the docker image build for e2e test is not mocked i.e. different from os_detector_test.go, I am not sure how to mock this.

agent/installer/os_detector.go
```diff
@@ -15,44 +15,69 @@ const (
...
+	if err := osd.DetectByHostnamectl(f); err != nil {
+		return "", err
+	}
+
+	return osd.normalizeOSName(), nil
+}
+
+// GetOSNameWithVersion detects and converts the detected os into format <os> <ver>
+func (osd *OSDetector) GetOSNameWithVersion(f func() (string, error)) (string, error) {
+	if osd.os != "" && osd.ver != "" {
```
sachinkumarsingh092, line 53: Just a small clarification: These `os` and `ver` values can also be fetched during reconciliation and that's why we do this optimization so as not to run `hostnamectl` again, right?

shubham14bajpai: Yes if the struct OSDetector is already contains those fields it skips calling `hostnamectl`. But the function declares the object each time it's called so the intention is not fulfilled. I guess we should declare the object at global level in the reconciler, right?

agent/reconciler/reconciler_test.go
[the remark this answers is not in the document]
shubham14bajpai: To make it generic we would have to identify the os in the test itself, which is same functionality we are testing in the case.

agent/installer/os_detector.go
sachinkumarsingh092: > I guess we should declare the object at global level in the reconcile

Yeah, makes sense!

agent/reconciler/host_reconciler.go
shubham14bajpai: Moved it back to registration.

shubham14bajpai:
Will create a new PR for the same.

PR #541, opened 2022-05-10
Create kubeconfig in bootstrap token flow

### LINKED ISSUE #518: Implement create kubeconfig for Bootstrap Token flow
With Host CSR approved, the Bootstrap authenticator should create a newer `kubeconfig` which will then be used by the host agent reconciler.

dharmjit:
Added changes for the creation of `kubeconfig` if the CSR is approved. This PR also introduces refactoring of the existing implementation. `BootstrapAuthenticator` controller is removed as the logic could be better handled in the `csr.go` code itself. The flow will wait till the timeout for certificate approval and if approved before the timeout will create `kubeconfig`. Some motivation is taken from the `kubelet bootstrapping` code
and client-go APIs are used as well.

Fixes #518


agent/authenticator/authenticator_suite_test.go
```diff
@@ -1,112 +0,0 @@
-// Copyright 2022 VMware, Inc. All Rights Reserved.
```
sachinkumarsingh092, line 1: All the `agent/authenticator/*` are removed because the Admission controller has required functionality, did I get that right?

dharmjit: The same is added now in `agent/main.go` and `agent/registration/csr.go`

agent/registration/csr.go
```diff
@@ -4,93 +4,133 @@
...
-		},
+	csrPemBlock := &pem.Block{
+		Type:  cert.CertificateRequestBlockType,
+		Bytes: csrData,
 	}
-	return privateKey, csr, nil
+	return pem.EncodeToMemory(csrPemBlock), nil
+}
+
+func LoadRESTClientConfig(bootstrapKubeconfig string) (*restclient.Config, error) {
```
sachinkumarsingh092, line 89: Would be good to add a small description for public functions.

agent/host_agent_test.go
```diff
@@ -557,8 +557,11 @@ var _ = Describe("Agent", func() {
 		})
 
 		AfterEach(func() {
-			// TODO: delete all CSR
-			// Expect(k8sClient.DeleteAllOf(ctx, &certv1.CertificateSigningRequest{}, client.MatchingFields{"metadata.name": fmt.Sprintf(registration.ByohCSRNameFormat, hostName)})).ShouldNot(HaveOccurred())
+			var list certv1.CertificateSigningRequestList
```
anusha94, line 560:
```suggestion
			var csrList certv1.CertificateSigningRequestList
```

agent/registration/csr.go
```diff
@@ -4,93 +4,133 @@
...
 )
 
 const (
-	KeySize           = 2048
-	ExpirationSeconds = 86400 * 365
-	ByohCSROrg        = "byoh:hosts"
-	ByohCSRCNFormat   = "byoh:host:%s"
-	ByohCSRNameFormat = "byoh-csr-%s"
+	KeySize            = 2048
+	ExpirationSeconds  = 86400 * 365
```
anusha94, line 28: A comment saying 1 day / 1 hour will be easier to read.

agent/main.go
```diff
@@ -258,3 +245,36 @@ func main() {
...
+	logger.Info("creating host csr")
+	bootstrapClientConfig, err := registration.LoadRESTClientConfig(bootstrapKubeConfig)
+	if err != nil {
+		return err
+	}
+	bootstrapClient, err := clientset.NewForConfig(bootstrapClientConfig)
+	if err != nil {
+		return err
+	}
+	byohCSR := registration.ByohCSR{BootstrapClient: bootstrapClient}
```
anusha94, line 262: Does this do the job of what we had originally intended for the bootstrap authenticator? As in wait / reconcile until the CSR is approved?

What happens in case of host agent restart?

dharmjit: > As in wait / reconcile until the CSR is approved?

yes, `csr.WaitForCertificate` will wait for 60 mins before timing out.

> What happens in case of host agent restart?

 `csr.RequestCertificate` takes care of the creation of the CSR and it skips creation if the CSR already exists. Currently, this check will fail on https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/util/certificate/csr/csr.go#L82 due the key mismatch. This will get handled with the private key being persisted across restarts.

agent/main.go
```diff
@@ -258,3 +245,36 @@ func main() {
...
+	logger.Info("Waiting for client certificate to be issued")
+	certData, err := csr.WaitForCertificate(ctx, bootstrapClient, reqName, reqUID)
+	if err != nil {
+		return err
+	}
+	keyData, err := keyutil.MarshalPrivateKeyToPEM(byohCSR.PrivateKey)
+	if err != nil {
+		return err
+	}
+	return registration.WriteKubeconfigFromBootstrapping(bootstrapClientConfig, "~/.kube/config", string(certData), string(keyData))
```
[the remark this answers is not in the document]
dharmjit: >Maybe something like ~/.byoh/config?

makes sense.

agent/host_agent_test.go
```diff
@@ -542,7 +542,7 @@ var _ = Describe("Agent", func() {
 
 			runner = setupTestInfra(ctx, hostName, getKubeConfig().Name(), ns)
 			runner.CommandArgs["--feature-gates"] = "SecureAccess=true"
-
+			runner.CommandArgs["--bootstrap-kubeconfig"] = "/mgmt.conf"
```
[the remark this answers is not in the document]
dharmjit: Actually wanted to use the same `kubeconfig` for the bootstrap process as well. Otherwise, have to create another config, which could be handled in `e2e` case. WDYT?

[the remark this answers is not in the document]
anusha94: > same kubeconfig for the bootstrap process as well

Didn't get this. What bootstrap process are you referring to?

agent/main.go
anusha94: Thanks.

agent/registration/csr.go
dharmjit: fixed.

agent/host_agent_test.go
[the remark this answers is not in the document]
dharmjit: I mean we cannot rename `mgmt.conf` to `bootstrap.conf` or otherwise we have to create another kubeconfig. I guess we can make these refactor in the final changes to support only one flow.

agent/host_agent_test.go
dharmjit: fixed.

shubham14bajpai (approved):
LGTM

agent/host_agent_test.go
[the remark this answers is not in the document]
dharmjit: Let's continue this discussion for the next developments.

PR #550, opened 2022-05-18
fixed: K8sInstallerConfig CR are not getting deleted when byomachine is deleted

### LINKED ISSUE #547: Byoh hostinfo.architecture is not matching with the supported bundle name
Byoh `hostinfo.architecture`  is not matching with the supported bundle name.   

byoh agent is setting architecture using `runtime.GOARCH` which provide `amd64` but the bundle name support `x86-64` means, we need to convert `amd64` to `x86-64` in InstallerConfig controller when generating the installation script.

Part of issue: #467

### LINKED ISSUE #548: K8sInstallerConfig CR are not getting deleted when byomachine is deleted
First create a cluster with byomachine infraRef as K8sInstallerConfigTemplate.
Next try to delete the cluster, the k8sinstallerconfig CRs created byomachine controller won't be deleted as they are trying to lookup for the byomachine which owns them but getting byomachine not found error.
byomachine is not waiting for installer CR to be deleted first before removing its finalizer.

On deleting the byomachine, the k8sinstallerconfig ,owns by respective byomachine, should be delete as well.

Belongs to #467

mayur-tolexo:
If installer resource owner byomachine not found and delete timestamp is set then also doing the delete reconciliation

Fixes #548

### DIFF
--- a/controllers/infrastructure/k8sinstallerconfig_controller.go
+++ b/controllers/infrastructure/k8sinstallerconfig_controller.go
@@ -16,7 +16,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/pointer"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	"sigs.k8s.io/cluster-api/util"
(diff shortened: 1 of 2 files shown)


shubham14bajpai (approved):
LGTM. I suggestion: Can we remove the deletion of secret from `reconcileDelete`? Reason being that secret does not have any finalizers and the ownerRef of installerConfig so it should get the delete call automatically.

shubham14bajpai (approved):

controllers/infrastructure/k8sinstallerconfig_controller.go
```diff
@@ -258,28 +258,6 @@ func (r *K8sInstallerConfigReconciler) ByoMachineToK8sInstallerConfigMapFunc(o c
 func (r *K8sInstallerConfigReconciler) reconcileDelete(ctx context.Context, scope *k8sInstallerConfigScope) (reconcile.Result, error) {
 	logger := scope.Logger
 	logger.Info("Deleting K8sInstallerConfig")
-
```
dharmjit, line 261: Are we relying on the K8s garbage collection mechanism to clean this?

dharmjit (approved):
LGTM

controllers/infrastructure/k8sinstallerconfig_controller.go
mayur-tolexo: Yup, we have owner reference set in the secret created by installer.

shamsher31 (approved):

PR #551, opened 2022-05-19
fixed: Byoh hostinfo.architecture is not matching with the supported bundle name

### LINKED ISSUE #547: Byoh hostinfo.architecture is not matching with the supported bundle name

mayur-tolexo:
Byoh hostinfo.architecture is not matching with the supported bundle name.
**Solution:**
in installer, replaced the new architecture name with the old one to
match with the bundle name.  

**Extra changes**
- fixed download package script command as output directory need to be created before pulling the bundle.
- Executed firewall enable/disable command only if ufw command exists.

Fixes #547

### DIFF
--- a/common/installer/installer.go
+++ b/common/installer/installer.go
@@ -17,10 +17,20 @@ type K8sInstaller interface {
 	Uninstall() string
 }
 
+// archOldNameMap keeps the mapping of architecture new name to old name mapping
+var archOldNameMap = map[string]string{
+	"amd64": "x86-64",
+}
+
 // NewInstaller will return a new installer
 func NewInstaller(ctx context.Context, osDist, arch, k8sVersion string, downloader BundleDownloader) (K8sInstaller, error) {
+	bundleArchName := arch
+	// replacing the arch name to old name to match with the bundle name
+	if _, exists := archOldNameMap[arch]; exists {
+		bundleArchName = archOldNameMap[arch]
+	}
 	// normalizing os image name and adding arch
-	osArch := strings.ReplaceAll(osDist, " ", "_") + "_" + arch
+	osArch := strings.ReplaceAll(osDist, " ", "_") + "_" + bundleArchName
 
 	reg := installer.GetSupportedRegistry(nil)
 	if len(reg.ListK8s(osArch)) == 0 {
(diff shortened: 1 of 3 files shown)


common/installer/installer.go
```diff
@@ -17,10 +17,20 @@ type K8sInstaller interface {
...
 // NewInstaller will return a new installer
 func NewInstaller(ctx context.Context, osDist, arch, k8sVersion string, downloader BundleDownloader) (K8sInstaller, error) {
+	bundleArchName := arch
+	// replacing the arch name to old name to match with the bundle name
+	if _, exists := archOldNameMap[arch]; exists {
+		bundleArchName = archOldNameMap[arch]
+	}
 	// normalizing os image name and adding arch
-	osArch := strings.ReplaceAll(osDist, " ", "_") + "_" + arch
+	osArch := strings.ReplaceAll(osDist, " ", "_") + "_" + bundleArchName
```
shubham14bajpai, line 33: This still won't fix the problem as the regex are different for os_detector and host_registration. We need to remove the "LTS" from osDist here. Small comparison test 🙂 : https://go.dev/play/p/QQ4IjESzFDo

mayur-tolexo: I have added "LTS" in test cases and it's working fine.  We check for the os distribution using regex: `Ubuntu_20.04.*_x86-64`. So, this won't be a problem.

shubham14bajpai: Ohh then it is perfect 👍

shubham14bajpai (approved):
LGTM

shubham14bajpai (approved):

common/installer/internal/algo/ubuntu20_4k8s.go
```diff
@@ -123,7 +124,9 @@ BUNDLE_PATH=$BUNDLE_DOWNLOAD_PATH/$BUNDLE_ADDR
 swapon -a && sed -ri '/\sswap\s/s/^#?//' /etc/fstab
 
 ## enable firewall
-ufw enable
+if command -v ufw >>/dev/null; then
```
dharmjit, line 127: Is this is to check if the `ufw` executable exists?

mayur-tolexo: Yes, so that it works on docker host as well (as docker host don’t have ufw command)

dharmjit (approved):

PR #566, opened 2022-05-30
Fixed bugs in agent bootstrap kubeconfig flow

dharmjit:
- fixed the user home dir substitution for the config file
- added list permission for `byoh-csr-creator-clusterrole`
- added the cert/key under appropriate keys in kubeconfig

Few issues as highlighted above were discovered while manually testing the `host-agent` with the new bootstrap flow.

### DIFF
--- a/agent/registration/csr.go
+++ b/agent/registration/csr.go
@@ -11,6 +11,7 @@ import (
 	"encoding/pem"
 	"fmt"
 	"os"
+	"path/filepath"
 	"time"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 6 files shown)


agent/registration/csr.go
```diff
@@ -34,10 +35,11 @@ const (
 	ByohCSRCNFormat   = "byoh:host:%s"
 	ByohCSRNameFormat = "byoh-csr-%s"
 	TmpPrivateKey     = "byoh-client.key.tmp"
+	DefaultConfigPath = ".byoh/config"
```
anusha94, line 38: can't we define the default as `~/.byoh/config`?

dharmjit: That's what was originally there, but we need to use `os.UserHomeDir()` to fetch home dir. This is now being done in `GetBYOHConfigPath()` function.

anusha94 (approved):
lgtm

PR #572, opened 2022-05-30
Fix: Use go install instead of go get to install binaries

dharmjit:
Makefile function `go-get-tool` uses `go get` to install go executables which is deprecated in Go 1.17 and starting Go 1.18, it will not support installing executables. Ref - https://go.dev/doc/go-get-install-deprecation

This also updates `Kustomize` to `v4.5.2` as the lower versions don't support `go install` due to exclude directive in go.mod Ref - https://github.com/kubernetes-sigs/kustomize/issues/3618


anusha94 (approved):
lgtm

shubham14bajpai (approved):

PR #578, opened 2022-06-06
kubeconfig flag removed in the agent

### LINKED ISSUE #582: Remove kubeconfig flag
With the new agent bootstrap kubeconfig flow, the existing `--kubeconfig` flag for the Host-Agent has to be removed to only allow the secure way to communicate with the mgmt cluster. This will also require updating the tests.

Part of - #484

dharmjit:
This PR removes the `--kubeconfig` flag from the HostAgent to support only the bootstrap kubeconfig flow via `--bootstrap-kubeconfig` flag. Some RBACs are updated as well.
- e2e tests which were earlier using `--kubeconfig` are updated and will use `--bootstrap-kubeconfig` flag with valid bootstrap kubeconfig with only permissions to create CSR.
- host agent tests will use a pre-placed kubeconfig in the default path `~/.byoh/config`. Please note that this kubeconfig is the one with admin priveleges.

Fixes #582


test/e2e/docker_helper.go
```diff
@@ -202,10 +202,37 @@ func (r *ByoHostRunner) copyKubeconfig(config cpConfig, listopt types.ContainerL
...
+		Cmd:          []string{"sh", "-c", "echo ${HOME}"},
+	})
+	Expect(err).ShouldNot(HaveOccurred())
+	resp, err := r.DockerClient.ContainerExecAttach(r.Context, execCommand.ID, types.ExecStartCheck{})
+	Expect(err).ShouldNot(HaveOccurred())
+	defer resp.Close()
+	homeDir, err := resp.Reader.ReadString('\n')
+	Expect(err).ShouldNot(HaveOccurred())
+	homeDir = strings.TrimSuffix(homeDir, "\n")
+	config.destPath = strings.SplitAfterN(strings.TrimSpace(homeDir)+"/.byoh/config", "/", 2)[1] // nolint: gomnd
```
shubham14bajpai, line 223: Just curious why do we need to split here?

shubham14bajpai (dismissed):
LGTM

test/e2e/docker_helper.go
dharmjit: good catch, reading the container exec response in `resp.Reader.ReadString('\n')` has some special characters in the prefix which were not even logged in `fmt.Print`. The output I was expecting is `/root/`

shubham14bajpai: okay 👍  Can we have comment on the line to increase readability.

shubham14bajpai (approved):

sachinkumarsingh092 (approved):
LGTM

agent/host_agent_test.go
```diff
@@ -544,7 +518,7 @@ var _ = Describe("Agent", func() {
 			Expect(err).NotTo(HaveOccurred())
 
 			runner = setupTestInfra(ctx, hostName, getKubeConfig().Name(), ns)
-			runner.CommandArgs["--bootstrap-kubeconfig"] = "/mgmt.conf"
+			runner.CommandArgs["--bootstrap-kubeconfig"] = "/root/.byoh/config"
```
anusha94, line 521: I thought we would remove `/root` hard coding?

dharmjit: We have removed the hardcodings for kubeconfig container destination path[[here](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/pull/578/files#diff-86318bdbaa9a116307312ef026fe917b23a193098696192e43b30ea0b53c076cR195)] but this is the one test where we are explicitly passing the `--bootstrap-kubeconfig` flag. To remove the hardcoding here we have to do container Exec and fetch the User home dir.

shubham14bajpai (approved):

anusha94 (approved):
LGTM.
I'll refactor testutils.go as part of #586

PR #579, opened 2022-06-07
Add e2e test for installer flow

### LINKED ISSUE #468: Write an e2e test for the new installer contract
For the installer contract proposed in #467 , write new e2e test(s) supporting the workflow

shubham14bajpai:
Signed-off-by: Shubham <shbajpai@vmware.com>

This PR adds e2e test for the new [installer flow](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/main/docs/diagrams/installer-flow.png) of provisioning workload cluster. This flow will replace the current way of provisioning so keeping that in mind few choices were made while implementing the test and to avoid writing complex tests which might be removed soon:
 - another cluster template was added with installer changes
 - the current test was duplicated with 1 key difference of `--use-installer-controller` enabled via host agent
 - another step was added to github action to run installer test separately as the both cannot be run together due to difference in cluster template
 - to reuse the same provider.yaml file `sed` the template path before running installer test.

The above choices also makes it easier to remove the old test when we decide to deprecate the old flow, which I assume is pretty soon. We can just remove the old test & template and rename/restructure some new test & templates.

Fixes #468 

**Additional information**
I was facing the below error while testing the installer flow. Reversing the order of kubelet and kubeadm in the install script fixed it
```
+ dpkg --install /var/lib/byoh/bundles/projects.registry.vmware.com/cluster_api_provider_bringyourownhost/byoh-bundle-ubuntu_20.04.1_x86-64_k8s:v1.23.5/kubeadm.deb
1Selecting previously unselected package kubeadm.
G(Reading database ... 8761 files and directories currently installed.)
(Preparing to unpack .../kubeadm.deb ...
"Unpacking kubeadm (1.23.5-00) ...
�dpkg: dependency problems prevent configuration of kubeadm:
 kubeadm depends on kubelet (>= 1.19.0); however:
  Package kubelet is not installed.
dpkg: error processing package kubeadm (--install):
 dependency problems - leaving unconfigured
```

### DIFF
--- a/test/e2e/data/infrastructure-provider-byoh/v1beta1/templates/e2e/installer/cluster-with-kcp.yaml
+++ b/test/e2e/data/infrastructure-provider-byoh/v1beta1/templates/e2e/installer/cluster-with-kcp.yaml
@@ -1,0 +1,153 @@
+---
+apiVersion: cluster.x-k8s.io/v1beta1
+kind: Cluster
+metadata:
+  name: ${CLUSTER_NAME}
+  labels:
+    cni: ${CLUSTER_NAME}-crs-0
+    crs: "true"
+spec:
+  clusterNetwork:
+    services:
+      cidrBlocks: ["10.128.0.0/12"]
+    pods:
+      cidrBlocks: ["192.168.0.0/16"]
+    serviceDomain: "cluster.local"
+  infrastructureRef:
+    apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
+    kind: ByoCluster
+    name: ${CLUSTER_NAME}
+  controlPlaneRef:
+    kind: KubeadmControlPlane
+    apiVersion: controlplane.cluster.x-k8s.io/v1beta1
+    name: "${CLUSTER_NAME}-control-plane"
+---
+apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
+kind: ByoCluster
+metadata:
+  name: ${CLUSTER_NAME}
+spec:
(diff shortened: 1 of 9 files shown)


test/e2e/data/infrastructure-provider-byoh/v1beta1/templates/e2e/installer/cluster-with-kcp.yaml
```diff
@@ -0,0 +1,153 @@
...
+  version: ${KUBERNETES_VERSION}
+---
+apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
+kind: K8sInstallerConfigTemplate
+metadata:
+  name: ${CLUSTER_NAME}-control-plane
+spec:
+  template:
+    spec:
+      bundleRepo: "projects.registry.vmware.com/cluster_api_provider_bringyourownhost"
```
dharmjit, line 152: We could use only one syntax for value i.e. either without quotes or with quotes everywhere.

dharmjit (commented):
LGTM on a first pass. minor nit only. Do we have a successful run to link in the PR?

shubham14bajpai:
> LGTM on a first pass. minor nit only. Do we have a successful run to link in the PR?

The e2e check in this PR runs both the old and flow test.

sachinkumarsingh092:
We should also close #255 after merging this PR.

anusha94 (approved):
lgtm

dharmjit (approved):

PR #595, opened 2022-06-15
Update ByoHost Validating webhook

### LINKED ISSUE #514: Add a Dynamic Admission Controller to limit ByoHost access to Host Agent
Write a dynamic admission webhook that gates byohosts from performing Read, Update, Delete on CRDs other than its own. This provides more security around host agent accessing / modifying other ByoHost resources.

dharmjit:
This PR updates the existing `ByoHost` validating webhook for authorization rules. Unit tests are also added along with integration tests.
Authz rules
- Update op will be allowed for the ByoHost owner agent user or controller manager service account.
- Create op will be allowed for the resource owner agent user
- Reads/Deletes are allowed to any user.

Fixes #514

### DIFF
--- a/apis/infrastructure/v1beta1/byohost_webhook.go
+++ b/apis/infrastructure/v1beta1/byohost_webhook.go
@@ -5,7 +5,9 @@ package v1beta1
 
 import (
 	"context"
+	"fmt"
 	"net/http"
+	"strings"
 
 	v1 "k8s.io/api/admission/v1"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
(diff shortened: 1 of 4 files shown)


apis/infrastructure/v1beta1/byohost_webhook.go
```diff
@@ -20,23 +22,55 @@ type ByoHostValidator struct {
...
+	case v1.Delete:
+		response = v.handleDelete(&req)
+	default:
+		response = admission.Allowed("")
 	}
+	return response
+}
 
-	// TODO: verify if req.UserInfo.Username has rbac permission to update the byohost
+func (v *ByoHostValidator) handleCreateUpdate(req *admission.Request) admission.Response {
```
Ankitasw, line 44: Can we create two separate functions `handleCreate` and `handleUpdate` and then call `handleCreate` inside `handleUpdate` put the if clause meant for update case in the same func, such that we don't mix two functionalities in one function, wdyt?

apis/infrastructure/v1beta1/byohost_webhook_internal_test.go
```diff
@@ -0,0 +1,198 @@
+// Copyright 2021 VMware, Inc. All Rights Reserved.
```
Ankitasw, line 1: Are we following any naming conventions for the integration tests and unit tests files? Unit tests should be by default byohost_webhook_test.go, and if we are adding any other tests like integration tests, it should go under byohost_webhook_integration_test.go, but that's a nit picking. Just wanted to make sure we are consistent with the naming.

apis/infrastructure/v1beta1/byohost_webhook.go
dharmjit: Yes, that was my initial implementation but as some of the code was going to be common, I kept it like this no strong opinions though. With the method name `handleCreateUpdate`, I guess it's okay from the readability perspective and code is also not very complicated at the moment.

>and then call handleCreate inside handleUpdate

This will bring some indirection which might not be very readable as to why we are calling handleCreate.

apis/infrastructure/v1beta1/byohost_webhook_internal_test.go
dharmjit: The convention we are following is to test the public APIs(types, methods) i.e. BlackBox testing and hence both the unit and integration tests can go in package_test.go file. If in case we have to do WhiteBox testing i.e. to test the unexported APIs, we are creating test files as package_internal_test.go

Ankitasw:
LGTM

apis/infrastructure/v1beta1/byohost_webhook.go
```diff
@@ -20,23 +22,55 @@ type ByoHostValidator struct {
 	decoder *admission.Decoder
 }
 
+// To allow byoh manager service account to patch ByoHost CR
+const managerServiceAccount = "system:serviceaccount:byoh-system:byoh-controller-manager"
+
 // nolint: gocritic
```
shamsher31, line 28: Any reason to add `nolint` here?

apis/infrastructure/v1beta1/byohost_webhook_internal_test.go
```diff
@@ -0,0 +1,198 @@
...
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "host1",
+					Namespace: "default",
+				},
+				Spec: ByoHostSpec{},
+			}
+			byoHostRaw, err = json.Marshal(byoHost)
+			Expect(err).ShouldNot(HaveOccurred())
+		})
+		It("Should allow delete request from any user", func() {
```
shamsher31, line 185: We should also add a test case for the following check?
```
cannot delete ByoHost when MachineRef is assigned
```

apis/infrastructure/v1beta1/byohost_webhook.go
dharmjit: This was already there, but I will check what its required for.

apis/infrastructure/v1beta1/byohost_webhook_internal_test.go
dharmjit: yup good to include in the internal tests as well.

apis/infrastructure/v1beta1/byohost_webhook.go
dharmjit: This is there for below lint issue, which couldn't be worked upon as this is an implementation of `admission.Handler` Interface 
`apis/infrastructure/v1beta1/byohost_webhook.go:29:56: hugeParam: req is heavy (400 bytes); consider passing it by pointer (gocritic)`

apis/infrastructure/v1beta1/byohost_webhook_internal_test.go
dharmjit: added.

shamsher31 (approved):
LGTM 👍

apis/infrastructure/v1beta1/byohost_webhook_internal_test.go
shamsher31: Thanks

apis/infrastructure/v1beta1/webhook_suite_test.go
```diff
@@ -87,6 +91,25 @@ var _ = BeforeSuite(func() {
 	Expect(err).NotTo(HaveOccurred())
 	Expect(k8sClient).NotTo(BeNil())
 
+	// apply Custom RBAC
```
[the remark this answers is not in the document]
dharmjit: fixed in the last commit.

apis/infrastructure/v1beta1/byohost_webhook_internal_test.go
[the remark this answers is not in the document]
dharmjit: fixed in the last commit.

anusha94 (approved):

Issue #593, opened 2022-06-15
Add documentation for installer controller implementation 


mayur-tolexo:
**Describe the solution you'd like**
Add documentation for installer controller implementation

PR #599, opened 2022-06-16
Deprecate old installer flow

### LINKED ISSUE #580: Deprecate old installer flow
Remove the old installer flow in favour of the new one mentioned in #467

shubham14bajpai:

This Pr finishes the refactoring of moving the installer logic out of the agent. During the refactoring to avoid breaking changes both the in-tree and installer controller flow were supported. This following tasks were undertaken to deprecate the in-tree flow:

- the installation logic under agent/installer is removed
- all the checks for in-tree installer were removed from agent and the controllers
- a new dir installer was introduced and the existing common/installer logic was moved to it along with any dependencies from the agent/installer
- the e2e test for the old flow was removed and replaced by the new test
- all the existing unit tests were updated/removed to used only the new flow
- the cluster templates were updated to include k8sInstallerConfigTemplate

The above involved removal of lots of code and tests as well. This resulted in a significant drop in test coverage. This PR also 
adds some new test cases over different packages to maintain the existing test coverage.

Fixes #580 

**Additional information**


agent/cloudinit/file_writer.go
```diff
@@ -90,7 +90,7 @@ func (w FileWriter) WriteToFile(file *Files) error {
 		ownerFormatLen := 2
 
 		if len(owner) != ownerFormatLen {
-			return errors.Wrap(err, fmt.Sprintf("Invalid owner format '%s'", file.Owner))
+			return errors.New(fmt.Sprintf("Invalid owner format '%s'", file.Owner))
```
dharmjit, line 93:
```suggestion
			return fmt.Errorf("Invalid owner format '%s': %w ", file.Owner, err)
```

agent/cloudinit/file_writer_test.go
```diff
@@ -37,6 +37,11 @@ var _ = Describe("FileWriter", func() {
 		Expect(err).NotTo(HaveOccurred())
 	})
 
+	It("Should fail to create a directory if path is invalid", func() {
```
dharmjit, line 40:
```suggestion
	It("should fail to create a directory if path is invalid", func() {
```

agent/cloudinit/file_writer_test.go
```diff
@@ -97,4 +102,40 @@ var _ = Describe("FileWriter", func() {
 		Expect(string(buffer)).To(Equal(fileOriginContent + file.Content))
 
 	})
+
+	It("Should create and write to file with invalid owner format", func() {
```
dharmjit, line 106:
```suggestion
	It("should return error with invalid owner format", func() {
```

agent/cloudinit/file_writer_test.go
```diff
@@ -97,4 +102,40 @@ var _ = Describe("FileWriter", func() {
...
+		}
+
+		err := cloudinit.FileWriter{}.MkdirIfNotExists(workDir)
+		Expect(err).NotTo(HaveOccurred())
+
+		err = cloudinit.FileWriter{}.WriteToFile(&file)
+		Expect(err).To(MatchError("Invalid owner format 'root'"))
+	})
+
+	It("Should create and write to file with unknown owner", func() {
```
dharmjit, line 124:
```suggestion
	It("should return error with unknown owner", func() {
```

installer/installer_suite_test.go
```diff
@@ -11,7 +10,7 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-func TestInstaller(t *testing.T) {
+func TestCloudinit(t *testing.T) {
```
dharmjit, line 13: Should it be `TestInstaller`?

dharmjit (commented):
Some nits only in the first pass

installer/installer_suite_test.go
shubham14bajpai: Yes copy paste error 😅

agent/reconciler/host_reconciler.go
```diff
@@ -119,7 +111,7 @@ func (r *HostReconciler) reconcileNormal(ctx context.Context, byoHost *infrastru
 
 		if r.SkipK8sInstallation {
 			logger.Info("Skipping installation of k8s components")
-		} else if r.UseInstallerController {
+		} else {
 			if !conditions.IsTrue(byoHost, infrastructurev1beta1.K8sComponentsInstallationSucceeded) {
```
mayur-tolexo, line 115:
```suggestion
		} else if !conditions.IsTrue(byoHost, infrastructurev1beta1.K8sComponentsInstallationSucceeded) {
```

agent/reconciler/host_reconciler.go
```diff
@@ -268,7 +252,7 @@ func (r *HostReconciler) hostCleanUp(ctx context.Context, byoHost *infrastructur
 		}
 		if r.SkipK8sInstallation {
 			logger.Info("Skipping uninstallation of k8s components")
-		} else if r.UseInstallerController {
+		} else {
 			if byoHost.Spec.UninstallationScript == nil {
```
mayur-tolexo, line 256:
```suggestion
		} else if byoHost.Spec.UninstallationScript == nil {
```

mayur-tolexo (commented):
Few Nits, overall LGTM

agent/reconciler/host_reconciler.go
shubham14bajpai: The condition `if byoHost.Spec.UninstallationScript == nil` ends at line 256 while the outer else ends at 271 so this will not work

mayur-tolexo: line#271 else is a part of its line#245 if condition. Isn’t it?

shubham14bajpai: Yes I meant 270 if we do as suggested then line 257 to 269 will become deadcode

mayur-tolexo: ooh ok got it. Then there is no need for else condition here. WDYT?

mayur-tolexo: Ok, my bad. I got confused by multiple if conditions.

agent/cloudinit/file_writer.go
```diff
@@ -90,7 +90,7 @@ func (w FileWriter) WriteToFile(file *Files) error {
 		ownerFormatLen := 2
 
 		if len(owner) != ownerFormatLen {
-			return errors.Wrap(err, fmt.Sprintf("Invalid owner format '%s'", file.Owner))
+			return fmt.Errorf("invalid owner format '%s'", file.Owner)
```
Ankitasw, line 93: Any specific reason for this change? As I see Wrap is used at L77 as well, so we should be consistent about this change.

installer/registry.go
```diff
@@ -47,7 +49,7 @@ func (r *registry) AddBundleInstaller(os, k8sVer string, installer osk8sInstalle
 		panic(fmt.Sprintf("%v %v already exists", os, k8sVer))
 	}
 
-	r.osk8sInstallerMap[os][k8sVer] = installer
+	r.osk8sInstallerMap[os][k8sVer] = empty
```
Ankitasw, line 52: is this required? as we are already initializing `osk8sInstallerMap` in `newRegistry` func(`make(osk8sInstallerMap)` will create an empty map), which is called before `AddBundleInstaller`

Ankitasw (commented):
Few nits, otherwise LGTM

agent/cloudinit/file_writer.go
shubham14bajpai: This was a bug which surfaced when we added test for this check. There is no error before this to wrap it.

installer/registry.go
shubham14bajpai: The make() will give us an empty map with no keys or values. Just to add keys to this support matrix we use empty interfaces. Example: https://go.dev/play/p/_MiQobRR64o

shubham14bajpai:
> Few nits, otherwise LGTM. It would also be worth checking if documentation is affected due to these changes, like getting-started.md or other dev docs.

Good catch 👍, removed the outdated documentation and there is a PR #593 for adding documentation for the new flow.

agent/cloudinit/file_writer_test.go
```diff
@@ -97,4 +97,40 @@ var _ = Describe("FileWriter", func() {
 		Expect(string(buffer)).To(Equal(fileOriginContent + file.Content))
 
 	})
+
+	It("should return error with invalid owner format", func() {
+		filePermission := 0777
```
shamsher31, line 102: Why the file permission is `777` and not `755`. The agent scripts will only be executed by the owner or anyone?

agent/cloudinit/cloudinit_integration_test.go
```diff
@@ -143,6 +156,13 @@ runCmd:
 		Expect(string(fileContents)).To(Equal(replacedFileContent))
 	})
 
+	It("should return error for invalid template content", func() {
+		cloudInitScript := "chcdslaxas;llc;"
```
anusha94, line 160: I understand the intention here, still can we change it to `"invalid-template"`, or `"foobar"`

agent/cloudinit/cloudinit_integration_test.go
```diff
@@ -60,6 +60,19 @@ runCmd:
...
 	})
 
+	It("should return error if execute commands fails", func() {
+		fileName := path.Join(workDir, "file-1.txt")
+		fileOriginContent := "some-content-8"
+		cloudInitScript := fmt.Sprintf(`write_files:
+- path: %s
+content: %s
+runCmd:
+- (exit 1)`, fileName, fileOriginContent)
```
anusha94, line 70: This test case is smart by using `exit 1` to fail the run cmd directive.
If the idea is to make Execute fail then even invalid script like `cloudInitScript := "foo"` should suffice, right?

agent/cloudinit/cloudinit_integration_test.go
shubham14bajpai: Yeahh makes it more readable 👍

agent/cloudinit/cloudinit_integration_test.go
shubham14bajpai: Yes

agent/cloudinit/file_writer_test.go
```diff
@@ -97,4 +97,38 @@ var _ = Describe("FileWriter", func() {
...
+		}
+
+		err := cloudinit.FileWriter{}.MkdirIfNotExists(workDir)
+		Expect(err).NotTo(HaveOccurred())
+
+		err = cloudinit.FileWriter{}.WriteToFile(&file)
+		Expect(err).To(MatchError("invalid owner format 'root'"))
+	})
+
+	It("should return error with unknown owner", func() {
```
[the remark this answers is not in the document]
shubham14bajpai: There is use of chown to change the owner in that function. I was getting the error `chown /tmp/file_writer_ut2246122839/file1.txt: operation not permitted` as the test are not run with permissions, still not sure how we can test this. Also this PR is to deprecate the installer flow, these tests were added just to avoid decrease in current coverage. I can take up tests in another PR 🙂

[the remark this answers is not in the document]
shamsher31: > I can take up tests in another PR


Make sense 👍

shamsher31 (approved):

Ankitasw (approved):

PR #606, opened 2022-06-21
Fix: Read kubeconfig first even with --bootstrap-kubeconfig flag

### LINKED ISSUE #598: Ignore `--bootstrap-kubeconfig` flag when config file already present

When we restart byo host agent, and if we passed `--bootstrap-kubeconfig` flag during the restart, then our controller fails with error message:
Screenshot 2022-06-15 at 5 14 20 PM

This happens because when we pass `--bootstrap-kubeconfig` flag again, then it generates the new CSR which is validated against old private key, which throws compatibility issue

To avoid this issue, we should ignore the `--bootstrap-kubeconfig` flag if config is already present in its path.

**Environment:**

dharmjit:
In the exisitng implementation, If the user passes the `--bootstrap-kubeconfig` in repeated invocations of the agent in the same host, the code flow is to create CSR again, but the k8s API `k8s.io/client-go/util/certificate/csr.RequestCertificate` tries to use the existing CSR and it will fail due to Certificate Key mismatch.  With this PR, `kubeconfig` if existing as `~/.byoh/config` will be used even with the `--bootstrap-kubeconfig` flag passed.

Fixes #598

### DIFF
--- a/agent/host_agent_test.go
+++ b/agent/host_agent_test.go
@@ -517,12 +517,10 @@ var _ = Describe("Agent", func() {
 			Expect(err).NotTo(HaveOccurred())
 
 			runner = setupTestInfra(ctx, hostName, getKubeConfig().Name(), ns)
-			runner.CommandArgs["--bootstrap-kubeconfig"] = "/root/.byoh/config"
+			runner.CommandArgs["--bootstrap-kubeconfig"] = "/root/config"
 			byoHostContainer, err = runner.SetupByoDockerHost()
 			Expect(err).NotTo(HaveOccurred())
 
-			output, _, err = runner.ExecByoDockerHost(byoHostContainer)
-			Expect(err).NotTo(HaveOccurred())
 			// Clean for any CSR present
 			var csrList certv1.CertificateSigningRequestList
 			Expect(k8sClient.List(ctx, &csrList)).ShouldNot(HaveOccurred())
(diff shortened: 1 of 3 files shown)


shivi28:
LGTM 👍

agent/main.go
```diff
@@ -149,15 +150,16 @@ func main() {
 		logger.Error(err, "could not determine hostname")
 		return
 	}
-
+	_, err = os.Stat(registration.GetBYOHConfigPath())
```
shamsher31, line 153: Should we also check for err?

agent/host_agent_test.go
```diff
@@ -561,7 +563,57 @@ var _ = Describe("Agent", func() {
 				return false
 			}, time.Second*2).Should(BeTrue())
 		})
+		It("should skip bootstrap kubeconfig flow if the ~/.byoh/config exists", func() {
```
mayur-tolexo, line 566: You have created `${HOME}/.byoh` directory but didn't clean up

mayur-tolexo (commented):
Small nit, overall LGTM

agent/host_agent_test.go
dharmjit: We are creating this inside container and the container gets destroyed after each test.

agent/main.go
dharmjit: We cannot exit if there are any errors in this. Do you mean to log it or something else? We are checking errors.Is(err, os.ErrNotExist) for the bootstrap kubeconfig flow

shamsher31: > We are checking errors.Is(err, os.ErrNotExist) for the bootstrap kubeconfig flow

This is fine

> Do you mean to log it or something else?

Yes, if `os.State` fails we should at least log an error?

dharmjit: Logs don't seem very nice
```
E0704 07:00:33.307968  517472 main.go:155]  "msg"="could not lookup kubeconfig file" "error"="stat /home/kokoni/.byoh/config: no such file or directory"  
I0704 07:00:33.308059  517472 main.go:236]  "msg"="initiated bootstrap kubeconfig flow" 
I0704 07:00:33.327298  517472 csr.go:78]  "msg"="CSR request created"  
I0704 07:00:33.327327  517472 csr.go:82]  "msg"="waiting for client certificate to be issued" 
```

mayur-tolexo (approved):

shamsher31 (approved):

PR #633, opened 2022-07-05
Webhook for bootstrapkubeconfig

### LINKED ISSUE #605: Add validation webhook for BootstrapKubeconfig API type
Per discussion on PR #546 , 

`BootstrapKubeconfig` type has 2 required fields in its `Spec` - `APIServer` and `CAData`.
- `APIServer` is the address of the kubernetes cluster and should be of the format `https://<host>:<port>`

- `CAData` should be a valid PEM encoded certificate authority data

Add validation webhooks to verify the sanity of these fields so that the user will be informed well in advance if the CR they are creating is of the incorrect format.

anusha94:
This PR adds a new validation webhook for BootstrapKubeconfig object
For a create / update operation,
- rejects if APIServer is empty
- rejects if APIServer is not of the form `https://hostname:port`
- rejects if CertificateAuthorityData is empty
- rejects if CertificateAuthorityData is not PEM encoded

Fixes #605

### DIFF
--- a/apis/infrastructure/v1beta1/bootstrapkubeconfig_webhook.go
+++ b/apis/infrastructure/v1beta1/bootstrapkubeconfig_webhook.go
@@ -1,0 +1,100 @@
+// Copyright 2022 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package v1beta1
+
+import (
+	b64 "encoding/base64"
+	"encoding/pem"
+	"net/url"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	ctrl "sigs.k8s.io/controller-runtime"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+)
+
+// log is for logging in this package.
+var bootstrapkubeconfiglog = logf.Log.WithName("bootstrapkubeconfig-resource")
+
+// APIServerURLScheme is the url scheme for the APIServer
+const APIServerURLScheme = "https"
+
+func (r *BootstrapKubeconfig) SetupWebhookWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewWebhookManagedBy(mgr).
+		For(r).
+		Complete()
+}
+
+//+kubebuilder:webhook:path=/validate-infrastructure-cluster-x-k8s-io-v1beta1-bootstrapkubeconfig,mutating=false,failurePolicy=fail,sideEffects=None,groups=infrastructure.cluster.x-k8s.io,resources=bootstrapkubeconfigs,verbs=create;update,versions=v1beta1,name=vbootstrapkubeconfig.kb.io,admissionReviewVersions=v1
+
+var _ webhook.Validator = &BootstrapKubeconfig{}
+
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
+func (r *BootstrapKubeconfig) ValidateCreate() error {
+	bootstrapkubeconfiglog.Info("validate create", "name", r.Name)
+
+	if err := r.validateAPIServer(); err != nil {
+		return err
+	}
+
+	if err := r.validateCAData(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
+func (r *BootstrapKubeconfig) ValidateUpdate(old runtime.Object) error {
+	bootstrapkubeconfiglog.Info("validate update", "name", r.Name)
+
+	if err := r.validateAPIServer(); err != nil {
+		return err
+	}
+
+	if err := r.validateCAData(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// ValidateDelete implements webhook.Validator so a webhook will be registered for the type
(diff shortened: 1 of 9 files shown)


apis/infrastructure/v1beta1/bootstrapkubeconfig_webhook.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+func (r *BootstrapKubeconfig) validateAPIServer() error {
+	if r.Spec.APIServer == "" {
+		return field.Invalid(field.NewPath("spec").Child("apiserver"), r.Spec.APIServer, "APIServer field cannot be empty")
+	}
+
+	// this is a basic check to see if the APIServer is of the format https://hostname:port
+	// will add more regex if needed
+	if len(strings.Split(r.Spec.APIServer, ":")) != APIServerSplitStringLength {
```
dharmjit, line 80: Could we use something from the standard library? this maybe https://pkg.go.dev/net/url#Parse

anusha94: yup. thanks. much better.

mayur-tolexo (dismissed):
LGTM

apis/infrastructure/v1beta1/bootstrapkubeconfig_webhook.go
```diff
@@ -0,0 +1,100 @@
...
+
+	if err := r.validateCAData(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// ValidateDelete implements webhook.Validator so a webhook will be registered for the type
+func (r *BootstrapKubeconfig) ValidateDelete() error {
```
shivi28, line 65: We can remove this function if there is no implementation. Wdyt?

apis/infrastructure/v1beta1/bootstrapkubeconfig_webhook.go
```diff
@@ -0,0 +1,100 @@
...
+	return nil
+}
+
+func (r *BootstrapKubeconfig) validateAPIServer() error {
+	if r.Spec.APIServer == "" {
+		return field.Invalid(field.NewPath("spec").Child("apiserver"), r.Spec.APIServer, "APIServer field cannot be empty")
+	}
+
+	parsedURL, err := url.Parse(r.Spec.APIServer)
+	if err != nil || parsedURL.Host == "" || parsedURL.Scheme != APIServerURLScheme || parsedURL.Port() == "" {
```
shivi28, line 78: Would you like to separate it out in separate function something like `func IsURLValid()` as this contains multiple checks

apis/infrastructure/v1beta1/bootstrapkubeconfig_webhook_test.go
```diff
@@ -0,0 +1,201 @@
...
+		testServerWithoutHostname   = "https://test-server"
+		testServerWithoutPort       = "https://test.com"
+		testServerValid             = "https://abc.com:1234"
+		testCADataEmpty             = ""
+		testCADataInvalid           = "test-ca-data"
+		testPEMDataInvalid          = b64.StdEncoding.EncodeToString([]byte(testCADataInvalid))
+	)
+	Context("When BootstrapKubeconfig gets a create request", func() {
+
+		It("should reject the request if APIServer field is empty", func() {
```
shivi28, line 36: Can add a test case if `url.Parse(r.Spec.APIServer)` returns an error

shivi28 (changes requested):

apis/infrastructure/v1beta1/bootstrapkubeconfig_webhook.go
anusha94: All 3 functions (`ValidateCreate`, `ValidateUpdate` and `ValidateDelete`) are needed, even though empty. Because are implementing the [Validator interface](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.11.2/pkg/webhook/admission#Validator)

apis/infrastructure/v1beta1/bootstrapkubeconfig_webhook_test.go
anusha94: Well, was looking for it. Turns out, [url.Parse](https://pkg.go.dev/net/url#Parse) returns empty or the characters as is in most cases - even for absurd strings. It is apparently due to "parsing ambiguities". I don't entirely follow what that means. But happy to add a case if I can find a string that causes `url.Parse` to return error

apis/infrastructure/v1beta1/bootstrapkubeconfig_webhook.go
anusha94: Although, I did add a test case for successful deletion now.
Thanks!

apis/infrastructure/v1beta1/bootstrapkubeconfig_webhook_test.go
mayur-tolexo: AFAIR, if you add space in protocol then it will give you an error eg. `htt p://test.com`

mayur-tolexo (dismissed):

shivi28:
lgtm

apis/infrastructure/v1beta1/bootstrapkubeconfig_webhook_test.go
anusha94: Ah! amazing. Adding it now :)

mayur-tolexo (approved):

shivi28 (approved):

PR #635, opened 2022-07-05
Fix Host Agent integration tests

### LINKED ISSUE #636: Host Agent Integration tests failing in CI
Host Agent tests are failing in CI for the Context `When the host agent is executed with --bootstrap-kubeconfig` with the below error

```
Unexpected error:
      <*fs.PathError | 0xc0009016b0>: {
          Op: "lstat",
          Path: "/tmp/mgmt.conf",
          Err: <syscall.Errno>0x2,
      }
      lstat /tmp/mgmt.conf: no such file or directory occurred
```

dharmjit:
Integration tests for Host Agent are failing as the `EnvTest` created kubeconfig is not copied to `/tmp/mgmt.conf`. This PR fixes that. 
Also to discover such issues in the future, we are also removing `/tmp/mgmt.conf` after each test execution

Fixes #636

### DIFF
--- a/agent/host_agent_suite_test.go
+++ b/agent/host_agent_suite_test.go
@@ -177,4 +177,11 @@ func cleanup(ctx context.Context, byoHostContainer *container.ContainerCreateCre
 			e2e.Showf("error removing log file %s: %v", agentLogFile, err)
 		}
 	}
+	_, err = os.Stat(e2e.TempKubeconfigPath)
+	if err == nil {
+		err = os.Remove(e2e.TempKubeconfigPath)
+		if err != nil {
+			e2e.Showf("error removing kubeconfig file %s: %v", e2e.TempKubeconfigPath, err)
+		}
+	}
 }
--- a/test/e2e/docker_helper.go
+++ b/test/e2e/docker_helper.go
@@ -27,7 +27,7 @@ import (
 
 const (
 	kindImage           = "byoh/node:e2e"
-	tempKubeconfigPath  = "/tmp/mgmt.conf"
+	TempKubeconfigPath  = "/tmp/mgmt.conf"
 	bootstrapKubeconfig = "/tmp/boostrap-kubeconfig"
 )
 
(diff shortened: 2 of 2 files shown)


agent/host_agent_suite_test.go
```diff
@@ -177,4 +177,11 @@ func cleanup(ctx context.Context, byoHostContainer *container.ContainerCreateCre
 			e2e.Showf("error removing log file %s: %v", agentLogFile, err)
 		}
 	}
+	_, err = os.Stat(e2e.TempKubeconfigPath)
```
shivi28, line 180: nit:
```
if _, err = os.Stat(e2e.TempKubeconfigPath); err == nil{
     if err = os.Remove(e2e.TempKubeconfigPath); err !=nil{
            e2e.Showf("error removing kubeconfig file %s: %v", e2e.TempKubeconfigPath, err)
     }
}
```

test/e2e/docker_helper.go
```diff
@@ -27,7 +27,7 @@ import (
 
 const (
 	kindImage           = "byoh/node:e2e"
-	tempKubeconfigPath  = "/tmp/mgmt.conf"
+	TempKubeconfigPath  = "/tmp/mgmt.conf"
```
shivi28, line 30: Why are we changing the case ? 🤔

shivi28 (commented):
One nit, else lgtm

test/e2e/docker_helper.go
sachinkumarsingh092: I guess this is to make it a global variable to be used in `agent/host_agent_suite_test.go`.

sachinkumarsingh092 (approved):

test/e2e/docker_helper.go
dharmjit: This issue in the previous PR was not caught as we were persisting the kubeconfig across tests. Now to remove this after each test we have to make this constant exportable.

agent/host_agent_suite_test.go
dharmjit: I do not have any strong opinions here, but for the first check, it is `err == nil` and that might be confusing with this one-liner err handling.

anusha94 (approved):

agent/host_agent_suite_test.go
shivi28: Makes sense

shivi28 (commented):
LGTM

PR #584, opened 2022-06-08
Advanced workflow for complete e2e

khannakshat7:
**What this PR does / why we need it**:
This workflow is needed when we want to run full e2e tests for the PR before merging.

**Additional information**
We have to add label `pr/run-all-e2e-test` to run all the e2e tests on the PR.

**Special notes for your reviewer**

I tried many different implementations-
1. Using flag `/test-full-e2e` to run e2e test but it failed as the event was triggered but was not updating in checks in the PR.
2. Using flag `/test-full-e2e` to create a workflow to add label to PR by Github bot and then running the e2e tests but it failed becoz a GitHub workflow cannot trigger another workflow.

PR #639, opened 2022-07-07
Cluster class upgrade e2e test

### LINKED ISSUE #620: [E2E] Upgrade test using cluster class
[A clear and concise description of what you want to happen.]
Related to #487 

[Miscellaneous information that will assist in implementing the feature.]

**Environment:**

Madhur97:
Adds e2e upgrade test using cluster class

Fixes #620 

The E2E test added in this PR has not been added to workflow as this upgrade test is not required to be run with every commit. This test can be run with other PRs after https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/pull/584 is completed. Successful run of this E2E test can be found [here](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/runs/7247188077?check_suite_focus=true).

### DIFF
--- a/test/e2e/clusterclass_upgrade_test.go
+++ b/test/e2e/clusterclass_upgrade_test.go
@@ -1,0 +1,186 @@
+// Copyright 2022 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// nolint: testpackage
+package e2e
+
+import (
+	"context"
+	"fmt"
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/client"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/utils/pointer"
+	"os"
+	"path/filepath"
+	"sigs.k8s.io/cluster-api/test/framework"
+	"sigs.k8s.io/cluster-api/test/framework/clusterctl"
+	"sigs.k8s.io/cluster-api/util"
+)
+
+var _ = Describe("Clusterclass upgrade test [K8s-upgrade]", func() {
+
+	var (
+		ctx                          context.Context
+		specName                     = "upgrade"
+		namespace                    *corev1.Namespace
+		cancelWatches                context.CancelFunc
+		clusterResources             *clusterctl.ApplyClusterTemplateAndWaitResult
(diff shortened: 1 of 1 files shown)


test/e2e/clusterclass_upgrade_test.go
```diff
@@ -0,0 +1,186 @@
+// Copyright 2022 VMware, Inc. All Rights Reserved.
```
anusha94, line 1: We can probably look at arranging our upgrade tests similar to what CAPI does - https://github.com/kubernetes-sigs/cluster-api/blob/main/test/e2e/cluster_upgrade_test.go
WDYT?

Madhur97: Yes, I had thought about this. But the issue is that the methods used in CAPI create a [custom](https://github.com/kubernetes-sigs/cluster-api/blob/0f2f85c4f2623f57f59a64888c66d1782f415d3f/test/e2e/cluster_upgrade.go#L113) namespace of their own while we need to pass the namespace in [command](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/pull/639/files#diff-91bfcaef6c403f73cf5d595e5f61366278e68e2f58fd1e728a5e00bbf26efaf4R79) args in BYO host runner. Due to this, the namespace used in docker container to run agent will be different from the resources created in upgrade test case.

anusha94: Can we pass this custom namespace to byohost runner?

Madhur97: We cannot pass the custom namespace to byohost runner since custom namespace is [randomly created](https://github.com/kubernetes-sigs/cluster-api/blob/0f2f85c4f2623f57f59a64888c66d1782f415d3f/test/e2e/cluster_upgrade.go#L113) at runtime inside methods of CAPI project and currently we do not have a way to control that namespace.

anusha94: @Madhur97 This is still in the `BeforeEach` right? What am I missing?

anusha94 (approved):
lgtm

test/e2e/clusterclass_upgrade_test.go
anusha94: Discussed offline. Makes sense to keep the tests separate

PR #706, opened 2022-09-15
Enable manual csr flow via clusterctl

### LINKED ISSUE #668: Add support for handling MANUAL_CSR_APPROVAL via environment variable/clusterctl
[A clear and concise description of what you want to happen.]

[Miscellaneous information that will assist in implementing the feature.]

**Environment:**

Madhur97:
This PR enables manual CSR flow via clusterctl. Now user can enable/disable manual csr approval by defining MANUAL_CSR_APPROVAL variable as enable/disable during clusterctl init. By default, it will remain disabled.

Fixes #668 

Have tested with MANUAL_CSR_APPROVAL as enable/disable using clusterctl.


main.go
```diff
@@ -132,8 +132,8 @@ func main() {
 		os.Exit(1)
 	}
 
-	// Set 'MANUAL_CSR_APPROVAL=true' to disable ByoAdmission controller. Now CSRs should be approved manually.
-	if os.Getenv("MANUAL_CSR_APPROVAL") != "true" {
+	// Set 'MANUAL_CSR_APPROVAL=enable' to disable ByoAdmission controller. Now CSRs should be approved manually.
+	if os.Getenv("MANUAL_CSR_APPROVAL") != "enable" {
```
anusha94, line 136: true / false is more intuitive IMO. Any reason for using enable / disable?

Madhur97: This change enables users to pass MANUAL_CSR_APPROVAL via clusterctl config. When we use true/false, it fails during conversion from unstructured object to concrete type [here](https://github.com/kubernetes/apimachinery/blob/master/pkg/runtime/converter.go#L363) because it assumes the value true/false as boolean while expected type is string.

shamsher31 (approved):

anusha94 (approved):

PR #739, opened 2022-11-01
Unit tests for certificate rotation

### LINKED ISSUE #656: Unit tests for certificate rotation
After #555, we need to write e2e tests for the same as the minimum time for certificate expiration is 10 minutes so this will be a slow test.

Madhur97:
This PR refactors the certificate rotation code to extract the logic into a separate method and adds unit tests for this method.
Fixes #656 

### DIFF
--- a/agent/host_agent_test.go
+++ b/agent/host_agent_test.go
@@ -31,6 +31,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	restclient "k8s.io/client-go/rest"
 	"k8s.io/klog/v2/klogr"
 	"sigs.k8s.io/cluster-api/util/conditions"
 	"sigs.k8s.io/cluster-api/util/patch"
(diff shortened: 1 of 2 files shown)


agent/host_agent_test.go
```diff
@@ -8,6 +8,7 @@ import (
 	"context"
 	"fmt"
 	"io/ioutil"
+	restclient "k8s.io/client-go/rest"
```
shivi28, line 11: Can we move this import from line 11 to somewhere between line 19-34

agent/host_agent_test.go
```diff
@@ -861,4 +862,80 @@ users:
...
...RGdnRVBBRENDQVFvQ2dnRUJBSjZvCnpZSnhVUnhnK2VHMkM4TzJOQXZ2OU9KZXIzL0lZTVBqcXh3cTNKMHZEbm02L3lHK0RQNzFxNWo5MzdlWmdMSncKZkI1YVJMMzFjdUw3N2RFUjZXc3UwVGRCRGN3MGJ6TWFEQ1lRT016TGdESnlrYmhlUmNtRkNDb1Z1MHVqaXlSbApyL3U2YllEZVcveWxmeTgxS1RJaWRJU0U1RUhoYjBlYWZyT0htbTY1aEVYSU5CbkF0WmJiUjZXaVk0NGt5SG1KCnh4bXBORGdHRHh4cVJQZ2ZCbDhrS2NkSDYvZSt1b3dPVlJtalY3a3ZsNzdtUi9uOEEvTUM5Ujc3enh0TmVCcXcKYzlmK3hmODRrMXlVT0xSZHM4M1JrNUFHZjB3Q2VjSmxMUmwwakNyNkJvSStZRm8xUWtMSzluQVJKZGtFS0xvbApaVFNkbHRodnBWd0MvTHdHeklzQ0F3RUFBYU5aTUZjd0RnWURWUjBQQVFIL0JBUURBZ0trTUE4R0ExVWRFd0VCCi93UUZNQU1CQWY4d0hRWURWUjBPQkJZRUZDUmh3bm1rcW9ncUZwS09ObmNUaFgvV2VvQTlNQlVHQTFVZEVRUU8KTUF5Q0NtdDFZbVZ5Ym1WMFpYTXdEUVlKS29aSWh2Y05BUUVMQlFBRGdnRUJBSnFWOHYxdC9kR0NrV0tXRzhxKwpWUTdqTHNZWkc5dUwyN2x6MVovYXFidUcxMURoemZCUHBxVDVSMHNkNXYwVFJUQ1lpK1JTVCs2UnB6Kzd1bytZCndWWUpBREwrcEE0UExTTGRGb1RUcHBFL1ZXcE5rbE0rZnhKa1U5cU5YUkRpN28xTzR4YWJOQUxRZWRwb3RWaVUKSFAxV3IxT3VNbDI0SW5vNkJYWU5CQ0JjT21mOWV3ZjV4Z2R5T2Z5eEh6WXI0dWJUb1dqT0hlaE1xUnQ1SUVqdApJRGQ2R0V0cG43VWVwdnFZNE50aU91UzFReUZodW13UGc2VmhtTjZZMFRrcjJrY2RYU3RtWVBxYzNGcU12RkRYClhDOTMzMkZrTjNxRWlNUFJvbUw0cDQvbHN2aUpLY0VpUG5iTTBCZzRWV1R0bjFqNWtwTVFNbC9TMU8rRkR6U20KWVlVPQotLS0tLUVORCBDRVJUSUZJQ0FURS0tLS0tCg==
```
shivi28, line 883: Can we use smaller string value for testing purpose?

agent/host_agent_test.go
```diff
@@ -861,4 +862,80 @@ users:
...
+			bootstrapKubeConf *os.File
+			err               error
+		)
+		BeforeEach(func() {
+			bootstrapKubeConf, err = ioutil.TempFile("", "bootstrap-kubeconfig")
+			Expect(err).NotTo(HaveOccurred())
+			bootstrapKubeConfig = bootstrapKubeConf.Name()
+		})
+		AfterEach(func() {
+			Expect(os.Remove(bootstrapKubeConf.Name())).ShouldNot(HaveOccurred())
```
mayur-tolexo, line 876:
```suggestion
			Expect(os.Remove(bootstrapKubeConfig)).ShouldNot(HaveOccurred())
```

agent/host_agent_test.go
```diff
@@ -861,4 +862,80 @@ users:
...
+		})
+		It("should return if certificate data is not valid", func() {
```
mayur-tolexo, line 878:
```suggestion
		It("should return error if certificate data is not valid", func() {
```

agent/host_agent_test.go
```diff
@@ -861,4 +862,80 @@ users:
...
+    client-certificate-data: aW5WYWxpZERhdGE=
+    client-key-data: aW5WYWxpZERhdGE=
+`)
+			_, err = bootstrapKubeConf.Write(testbootstrapKubeconfigInvalid)
+			Expect(err).NotTo(HaveOccurred())
+			var config *restclient.Config
+			config, err = registration.LoadRESTClientConfig(bootstrapKubeConfig)
+			Expect(err).NotTo(HaveOccurred())
+			err = certRotation(klogr.New(), "test-host", config)
+			Expect(err).ShouldNot(HaveOccurred())
```
mayur-tolexo, line 907: I'm not sure what all errors can occur in `certRotation` function but do you need to verify the error returned is the same as expected?

agent/main.go
```diff
@@ -246,34 +246,40 @@ func handleBootstrapFlow(logger logr.Logger, hostName string) error {
...
 
-		// Poll after every few seconds
-		time.Sleep(pollDuration)
+	totalTimeCert := cert.NotAfter.Sub(cert.NotBefore)
+
+	// if less than 20% time left, renew the certs.
+	// https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/proposals/20210222-kubelet-authentication.md#kubelet-authenticator-flow
+	if time.Now().After(cert.NotAfter.Add(totalTimeCert / -5)) {
+		logger.Info("certificate expiration time left is less than 20%, renewing")
+		if err = handleBootstrapFlow(logger, hostName); err != nil {
```
mayur-tolexo, line 276: You are not returning error returned from `handleBootstrapFlow` function call.

mayur-tolexo: ok, my bad. certRotation is called in an infinite loop after poll duration. This looks good. 👍

agent/host_agent_test.go
Madhur97: Existing code does not return error if certificate data is not valid (it just logs the failures). So this test also checks that the method should return without error.

agent/main.go
Madhur97: This PR refactors the existing code. As per the existing code, the error from `handleBootstrapFlow` function call is logged and it is not returned to the caller. I am not sure but perhaps this was done to prevent the byoh agent from crashing when there is some error during certificate rotation.

agent/host_agent_test.go
Madhur97: Existing code does not return error if certificate data is not valid (it just logs the failures). So this test also checks that the method should return without error.

agent/host_agent_test.go
```diff
@@ -861,4 +862,80 @@ users:
 			Expect(err.Error()).Should(ContainSubstring("kubeconfig generation failed: hostname is not valid"))
 		})
 	})
+	Context("When the certRotation func is called", func() {
+		var (
+			bootstrapKubeConf *os.File
+			err               error
+		)
+		BeforeEach(func() {
+			bootstrapKubeConf, err = ioutil.TempFile("", "bootstrap-kubeconfig")
```
dharmjit, line 871: nit: s/bootstrapKubeConf/kubeConfig/g as certRotation will be called once the Bootstrap flow is completed.

agent/main.go
```diff
@@ -246,34 +246,40 @@ func handleBootstrapFlow(logger logr.Logger, hostName string) error {
...
+		if err := certRotation(logger, hostName, config); err != nil {
 			return err
 		}
+		// Poll after every few seconds
+		time.Sleep(pollDuration)
+	}
+}
 
-		totalTimeCert := cert.NotAfter.Sub(cert.NotBefore)
+func certRotation(logger logr.Logger, hostName string, config *rest.Config) error {
```
dharmjit, line 257: Suggestion: We have extracted all of the code except for the loop in this function. Could you see if it's okay to move the for loop [here](https://github.com/vmware-tanzu/cluster-api-provider-bringyourownhost/blob/main/agent/main.go#L179)?

dharmjit (commented):
some nits only.

agent/host_agent_test.go
Madhur97: The restclient import was present in a different import group separated by a blank line, perhaps that's why goimports did not take care in this case.

agent/main.go
Madhur97: As the main method is already handling a lot of things, to avoid more complexity it would be better to keep the for loop in a method of its own.

dharmjit (approved):
LGTM! Thanks.

mayur-tolexo (approved):
LGTM Nice work 👍

shivi28:
LGTM

PR #905, opened 2025-07-04
Excluding /etc/cni/net.d from being cleaned up to avoid deleting non apps-infra networks

amol-cohesity:
Excludes /etc/cni/net.d from being cleaned up as we already handle it internally. 
Also the fact if we delete /etc/cni/net.d, it would also delete the other networks beyond scope of AppsInfra

Fixes #
ENG-518434

**Additional information**


amol-cohesity:
Created by mistake

PR #910, opened 2026-05-15
Eng 643015 preserve kubelet binary in hostagent error handling path

piyush-bhoi:
1. When kubeadm join/init fails during bootstrap, the hostagent's error-handling path calls resetNode() which removes /usr/bin/kubelet. However, on the next reconcile, K8sComponentsInstallationSucceeded is still True, so the installer is skipped ("install script already executed"), meaning no one copies kubelet back. 
2. This creates an infinite loop: join fails => kubelet deleted => retry skips install => join fails again (no kubelet).

This PR introduces a resetNodeMode enum to distinguish between two reset contexts:

1. **resetNodeModeBootstrapRetry** (bootstrap error path): skips kubelet binary removal, preserving /usr/bin/kubelet so the next reconcile can retry join without re-install.
**resetNodeModeHostCleanup** (full host teardown): still removes the kubelet binary, which is safe because K8sComponentsInstallationSucceeded is reset to False afterward, guaranteeing the installer re-runs.

Fixes ENG-643015 - https://jira.cohesity.com/browse/ENG-643015

**Additional information**

The fix was verified end-to-end on a live node (10.136.8.165):

1. Corrupted the bootstrap secret (controlPlaneEndpoint => unreachable IP) to force kubeadm init failure.
2. Confirmed the error path was hit ("error in bootstrapping k8s node" at host_reconciler.go:163).
3. Verified removeKubeletBinary=false was passed and "**Skipping kubelet binary removal for bootstrap retry reset**" was logged.
4. Confirmed /usr/bin/kubelet survived the error path (md5 unchanged).
5. Restored the bootstrap secret and confirmed the node re-bootstrapped successfully on the next reconcile without needing to re-run the installer.


piyush-bhoi:
This is mistakenly open from a wrong branch, will open one separate PR.

